PR #5, opened 2019-11-03
Ilya_Torch_itorch2001@gmail.com

IlyaTorch:
Iteration 1

Combine functions for parsing console arguments, for handling options,
for handling entries into classes using dataclasses module. Add type of accepted and returned
values of methonds. Command line points -h, --version, --json, --limit are ready.


HenadziStantchik:
Hello! Please update your branch:  fetch changes in our README.md and some renames we did yesterday.

final_task/rss_parser/rss_parser.py
```diff
@@ -0,0 +1,156 @@
...
+    arg_parser.add_argument("--verbose", action="store_true", help="Outputs verbose status messages")
+    arg_parser.add_argument("--limit", type=int, default=1, help="Limit news topics if this parameter provided")
+    arg_parser_args = arg_parser.parse_args()
+
+    def __post_init__(self):
+        self.source = self.arg_parser_args.source
+        self.status_version = self.arg_parser_args.version
+        self.status_json = self.arg_parser_args.json
+        self.status_verbose = self.arg_parser_args.verbose
+        self.limit = self.arg_parser_args.limit
```
HenadziStantchik, line 29: You don't really need separate class for this. You will have all parsed arguments already after line arg_parser_args = arg_parser.parse_args()

final_task/rss_parser/rss_parser.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+@dataclass
```
HenadziStantchik, line 32: You do not really need to use dataclass for this. Simple class will do. Dataclasses are meant to be used as compact data-containing objects with zero or almost absent method logic. Here we have full-fledged objects with variety of methods.

final_task/rss_parser/rss_parser.py
```diff
@@ -0,0 +1,156 @@
...
+        print("Link: ", self.article_link)
+
+    def print_summary(self) -> None:
+        print('\n')
+        print(self.summary)
+        print('\n')
+
+    def print_links(self) -> None:
+        print("Links:")
+        for i, link in enumerate(self.links):
```
HenadziStantchik, line 60: It is better to avoid one-letter variable names, even in such simple constructions.

final_task/rss_parser/rss_parser.py
```diff
@@ -0,0 +1,156 @@
...
+            self.summary = self.summary[: start_cut] \
+                            + f"[image {len(self.links) - 1}: " + alt + "]" \
+                            + f"[{len(self.links) - 1}] " \
+                            + self.summary[self.summary.find(">", start_cut + len("<img")) + 1:]
+
+        while self.summary.count('<'):
+            self.summary = self.summary[:self.summary.find('<')] + self.summary[self.summary.find('>') + 1:]
+
+
+@dataclass()
```
HenadziStantchik, line 84: Same as above, you do not really need to use dataclass here

final_task/rss_parser/rss_parser.py
```diff
@@ -0,0 +1,156 @@
...
+            "Feed": self.parsed.feed.title,
+            "Title": entry.title,
+            "Date": entry.date,
+            "Link": entry.article_link,
+            "Links": entry.links
+        }
+        return entry_dict
+
+
+def main():
```
HenadziStantchik, line 139: It is common practice to separate each class in its own module.

final_task/rss_parser/Handler.py
```diff
@@ -0,0 +1,60 @@
+import json
+from dataclasses import dataclass
+import feedparser
+
+from rss_parser.Entry import Entry
+
+
+@dataclass()
```
dzhigailo, line 8: It's great, that you have used dataclasses, but they do not suppose to contain logic. In this particular case, class Hadler should be a default class

PR #8, opened 2019-11-05
Ilya_Torch_itorch2001@gmail.com

IlyaTorch:
Iteration 1
    Options --version, --json, --limit, --verbose are ready. Add option --verbose. Split classes into files. Add description of project to file README.md.


final_task/README.md
```diff
@@ -1,3 +1,61 @@
...
+    "https://news.yahoo.com/wet-weekend-tropical-storm-warnings-131131925.html",
+    "http://l2.yimg.com/uu/api/res/1.2/Liyq2kH4HqlYHaS5BmZWpw--/YXBwaWQ9eXRhY2h5b247aD04Njt3PTEzMDs-/https://media.zenfs.com/en/ap.org/5ecc06358726cabef94585f99050f4f0"
+  ]
+}
+```
+The `--limit` argument also affects JSON generation.
+
+With the argument `--verbose` program prints all logs in stdout.
+
+Withe the argument `--version` program prints in stdout it's current version and complete it's work.
```
HenadziStantchik: You are supposed to write documentation yourself. Not to copy it from the task.

final_task/rss_reader/Entry.py
```diff
@@ -0,0 +1,77 @@
...
+class Entry:
+    """class for every article from http:link...link.rss"""
+
+    def __init__(self, title: str, date: str, article_link: str, summary: str, links: list):
+        self.title = title
+        self.date = date
+        self.article_link = article_link
+        self.summary = summary
+        self.links = links
+        self.parse_html()
```
HenadziStantchik: It is better to make this method to return parsed summary and write it like this:  
`self.summary = self.parse_html(summary)`  

Right now you assign `self.summary = summary` and after that modify it anyway. This assignment is a bit redundant

final_task/rss_reader/Entry.py
```diff
@@ -0,0 +1,77 @@
...
+        logging.info("Entry object created")
+
+    def print_title(self) -> None:
+        logging.info("function \"print_title\" started")
+
+        print("Title: ", self.title)
+
```
HenadziStantchik: Here and below: you do not really need empty lines here. Code will still be pretty readable without them

final_task/rss_reader/Entry.py
```diff
@@ -0,0 +1,77 @@
...
+        logging.info("function \"print_title\" finished")
+
+    def print_date(self) -> None:
+        logging.info("function \"print_date\" started")
+
+        print("Date: ", self.date)
+
+        logging.info("function \"print_date\" finished")
```
HenadziStantchik: Here, above and below. You can make these lines better (logging part). Think about actions you perform here:

1. You perform some logic (write logs) before performing function itself.
2. You perform function.
3. You perform some logic (write logs) after function.

We have a potential for a good decorator here. It is perfect example of decorator usage.

final_task/rss_reader/Handler.py
```diff
@@ -0,0 +1,90 @@
...
+        for num, entry in enumerate(self.gen_entries()):
+            if num == self.limit:
+                break
+            entry.print_title()
+            entry.print_date()
+            entry.print_link()
+            entry.print_summary()
+            entry.print_links()
+            if num == self.limit - 1:
+                break
```
HenadziStantchik: A question:  

Why do you need this here if you already have almost the same check a bit above?

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,38 @@
...
+
+def main():
+    arg_parser = argparse.ArgumentParser(description="Pure Python command-line RSS reader.")
+    arg_parser.add_argument("source", type=str, help="RSS URL")
+    arg_parser.add_argument("--version", action="store_true", help="Print version info")
+    arg_parser.add_argument("--json", action="store_true", help=" Print result as JSON in stdout")
+    arg_parser.add_argument("--verbose", action="store_true", help="Outputs verbose status messages")
+    arg_parser.add_argument("--limit", type=int, default=1, help="Limit news topics if this parameter provided")
+
+    arg_parser_args = arg_parser.parse_args()
```
HenadziStantchik: It is better to move this to a separate function

HenadziStantchik:
Please do NOT close this and open new pull request anymore, Our comments do not save and it is hard to track your progress

HenadziStantchik:
Here is some feedback about your application work:
1. If you to check version without specifying RSS url, you wil get:
`rss_reader.py: error: the following arguments are required: source`  
It should be that you do not need any RSS url for checking app version.
2. You got some problems with encoding. Check image below:

image

3. If limit is not specified you will get only 1 piece of news right now, while it should give you the whole available feed
4. Sometimes you do not get image link in "Links" (see the image below)

image

P.S. I used --limit 999 option to test last threee points.

final_task/README.md
IlyaTorch: Why? I did parser completely according to the task and all the options are listed in the task.

HenadziStantchik: But if you had inserted example of your program working you could have seen that there are some problems with encoding at least. We do not need to read about which arguments your application take because we know it anyway.  

We want you to describe how does your application work in the documentation and probably to list some peculiarities of your implementation, some explanations and etc.

final_task/rss_reader/Entry.py
IlyaTorch: fixed

final_task/README.md
IlyaTorch: Got it

final_task/rss_reader/Entry.py
IlyaTorch: fixed

final_task/rss_reader/Entry.py
IlyaTorch: fixed

final_task/rss_reader/Handler.py
IlyaTorch: I think that I need this condition because in the cicle: 
"for num, entry in enumerate(self.gen_entries()):
a new entry will be created and only after it I'll get False in this condition:
if num == self.limit:
    break 
i.e an additional object will be created.

final_task/rss_reader/rss_reader.py
IlyaTorch: fixed

IlyaTorch:
1), 3) -fixed
2) Is it necessary?
4) Img tag exists on this page and it has an attribute "alt" while it doesn't have an attribute "src". What do I need to do in this case?

HenadziStantchik:
> 2) Is it necessary?

Yes

> 4) Img tag exists on this page and it has an attribute "alt" while it doesn't have an attribute "src". What do I need to do in this case?

That is certainly strange. Did you visit the url? if there is no src then there should be no image available on the page.

P.S. In the future use labels. Check slack for more info.

final_task/rss_reader/Handler.py
HenadziStantchik: That is true, but why do you need `if` condition at the top then? you will never hit it before the lower one anyway

final_task/rss_reader/Handler.py
```diff
@@ -0,0 +1,73 @@
...
+    """class for handling different options: --version, --json, --limit Limit"""
+    @logging_decorator
+    def __init__(self, source: str, limit: int, version: float):
+        self.source = source
+        self.limit = limit
+        self.version = version
+        self.parsed = feedparser.parse(self.source)
+        logging.info("Handler object created")
+
+    """options of command line:"""
```
HenadziStantchik: Usually docstrings are ituated only at the beginning of the function/class/module. Here specifically it is better to make that a comment

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,28 @@
+import argparse
+from Handler import Handler
+from Logging import logging_decorator
+from Console_parse import get_arguments_from_console
```
HenadziStantchik: I do not see these files in this PR. Are they 3rd-party libraries? If so, please update requirements.txt file

final_task/rss_reader/Handler.py
IlyaTorch: If limit is equal for example 0 we'll leave the cycle and won't print information about an entry. Or limit is equal to 1 then we'll get the information about entry and leave the cycle because this condition:
 if num == self.limit - 1:
                break
will return true. If this condition would not exist, we would get one more object entry because it creates in the cicle:
"for num, entry in enumerate(self.gen_entries())
These 2 conditions are necessary to not to get one more extra object.
And I don't know what is more important in this case: not to get an extra object or more readable code.

final_task/rss_reader/rss_reader.py
IlyaTorch: I forgot to commit these files.

final_task/rss_reader/Handler.py
IlyaTorch: fixed

IlyaTorch: And I have a question about the point 4 above:
\See Photos of the SpeedKore Dodge Charger
this is a string from url "https://news.yahoo.com/rss/"
And this image isn't on the html version of this page. What can I do in this case?

IlyaTorch:
is:open label:question

final_task/rss_reader/Handler.py
HenadziStantchik: Well, that is still strange, but in this case, when you have no text and no image it seems logical to me to ignore this piece of news in the feed

final_task/rss_reader/Handler.py
HenadziStantchik: Now I get it. Then it is better to remove upper check and make it so that `--limit` must be > 0 if it was specified by user.

Upper check will trigger only in case if `limit == 0` and in any other case the cycle will break at the bottom check.

final_task/rss_reader/rss_reader.py
HenadziStantchik: Still do not see these files. There is nothing for me to review

HenadziStantchik:
Also please do not `request review` using site interface, because it triggers a bit different functionality and compells us to create feedbacks abouth your PR and either merge it or close it. Cancel review request please

final_task/rss_reader/Handler.py
IlyaTorch: In such articles text exists but img tags has no attribute src

IlyaTorch:
When can I cancel review request?

HenadziStantchik:
When launching your application in this way:
`python rss_reader.py https://news.yahoo,com/rss/`
I get an exception:
image
This bug is inconsistent because this appears not every application run.

final_task/rss_reader/Entry.py
```diff
@@ -0,0 +1,77 @@
...
+        for symbol in ['&iquest;', '&iexcl;', '&laquo;', '&raquo;', '&lsaquo;', '&rsaquo;', '&quot;', '&lsquo;',
+                    '&rsquo;', '&ldquo;', '&rdquo;', '&sbquo;', '&bdquo;', '&sect;', '&para;', '&dagger;',
+                    '&Dagger;', '&bull;', '&mdash;', '&ndash;', '&hellip;', '&nbsp;'
+                    ]:
+            while summary.count(symbol):
+                summary = summary[:summary.find(symbol)] + '"' + summary[summary.find(symbol) + len(symbol):]
+        # decode ASCII codes:
+        while summary.count("&#"):
+            code = summary[summary.find("&#") + 2: summary.find(";")]
+            summary = summary[:summary.find("&#")] + chr(int(code)) + summary[summary.find("&#") + 3 + len(code):]
```
HenadziStantchik: Try to look into some string standart functions like encode(), decode() for fixing encoding. It should be possible to use them to solve encoding problem in a more fast and elegant manner

IlyaTorch:
It's strange because when I launch application using: 
`python rss_reader.py https://news.yahoo,com/rss/`
command I get only my expect-exception.
image

final_task/rss_reader/Entry.py
IlyaTorch: I couldn't find a way how to do it using encode() or decode() functions so what I used function html.unescape() for it.

HenadziStantchik:
Looks good so far. Keep it up

IlyaTorch:
I have a problem with wrapping package:
`python setup.py sdist`
`cd dist`
`pip install rss-reader-1.0.tar.gz`
and when I try to run it:
`rss-reader -h`
I get the following error:
`ModuleNotFoundError: No module named 'Handler'`
But when I run the project from directory where it was installed I don't get errors
What could be the problem?
image

HenadziStantchik:
> I have a problem with wrapping package:
> `python setup.py sdist`
> `cd dist`
> `pip install rss-reader-1.0.tar.gz`
> and when I try to run it:
> `rss-reader -h`
> I get the following error:
> `ModuleNotFoundError: No module named 'Handler'`
> But when I run the project from directory where it was installed I don't get errors
> What could be the problem?

This means either you wrapped it not exatly in correct way or the problem is that package name and module name are the same (rss_reader). Try to use sys.path.append to bypass this

IlyaTorch:
Where can I use sys.path.append?

final_task/README.md
```diff
@@ -1,3 +1,60 @@
...
+Link:  https://news.yahoo.com/graham-trump-ukraine-incoherent-quid-pro-quo-192210175.html
+
+
+[image 1: Graham now says Trump's Ukraine policy was too 'incoherent' for quid pro quo][1] A day after saying he wouldn’t bother to read the testimony, Sen. Lindsey Graham now says he did read it, and his conclusion is that the Trump administration’s Ukraine policy was too "incoherent" for it to have orchestrated the quid pro quo at the heart of the impeachment inquiry.
+
+
+Links:
+[0]  https://news.yahoo.com/graham-trump-ukraine-incoherent-quid-pro-quo-192210175.html (link)
+[1]  http://l2.yimg.com/uu/api/res/1.2/aWhGys7_IW5qIjKaiJpPfg--/YXBwaWQ9eXRhY2h5b247aD04Njt3PTEzMDs-/https://media-mbst-pub-ue1.s3.amazonaws.com/creatr-uploaded-images/2019-11/5527ffe0-00ca-11ea-9f7d-d1e736c1315d (image)
+```
```
HenadziStantchik: Please add instructions on how to launch your app using packaging tools on a clear machine here.

HenadziStantchik:
Looks good so far, proceed to the next iteration. Keep in mind that `--limit` should work with `--date`

final_task/rss_reader/Entry.py
```diff
@@ -0,0 +1,67 @@
+import logging
+from Logging import logging_decorator
+import html
+
+
+class Entry:
+    """class for every article from http:link...link.rss"""
+    @logging_decorator
+    def __init__(self, title: str = "", date: str = "", article_link: str = "", summary: str = "", links: list = []):
```
dzhigailo: During our course we have explained why is it bad to set mutable objects as default arguments. Please, fix it

final_task/rss_reader/Handler.py
```diff
@@ -0,0 +1,69 @@
...
+    # options of command line:
+    @logging_decorator
+    def option_version(self) -> None:
+        print("version ", self.version)
+
+    @logging_decorator
+    def option_json(self) -> None:
+        for num, entry in enumerate(self.gen_entries()):
+            self.print_to_json(self.convert_to_dict(entry))
+            if num == self.limit - 1:
```
dzhigailo: Insted of checking the limit on each iteration, you can define the number of iteration to be no more than a limit. For example:
```
for item in item_list[:limit]:
 ...
```
In this case you won't need `break` and `if`

final_task/rss_reader/Handler.py
```diff
@@ -0,0 +1,69 @@
...
+    @logging_decorator
+    def option_default(self) -> None:
+        self.print_feed()
+        for num, entry in enumerate(self.gen_entries()):
+            entry.print_title()
+            entry.print_date()
+            entry.print_link()
+            entry.print_summary()
+            entry.print_links()
+            if num == self.limit - 1:
```
dzhigailo: Same as above

final_task/rss_reader/Handler.py
dzhigailo: Same above

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+@logging_decorator
+def main():
+    arg_parser_args = get_arguments_from_console()
+    version = 1.0
+
+    if arg_parser_args.verbose:
+        logging.basicConfig(level=logging.INFO)
+    if arg_parser_args.limit <= 0:
```
dzhigailo: If limit<=0, you should raise exception, because it's a malformed input argument

IlyaTorch: fixed

final_task/rss_reader/Handler.py
IlyaTorch: In my case method gen_entries() is a generator and I can't use the following construction:
`
for item in item_list[:limit]:
`
Can I use the following construction:
`
for entry in islice(self.gen_entries(), 0, self.limit):
`
instead of your example?

final_task/rss_reader/Entry.py
IlyaTorch: fixed

final_task/rss_reader/Handler.py
dzhigailo: I'm not sure if you really need generator in here:)

final_task/rss_reader/cache.json
```diff
@@ -0,0 +1,122 @@
...
+    "DateInt": "20191112",
+    "Date": "Tue, 12 Nov 2019 13:46:43 -0500",
+    "Link": "https://news.yahoo.com/could-president-trump-be-impeached-and-removed-from-office-but-still-reelected-184643831.html",
+    "Summary": "[image 1: Could President Trump be impeached and removed from office \u2014 but still reelected?][1] What happens when a presidential impeachment inquiry runs into a presidential election year? The United States in uncharted territory.",
+    "Links": [
+      "https://news.yahoo.com/could-president-trump-be-impeached-and-removed-from-office-but-still-reelected-184643831.html",
+      "http://l2.yimg.com/uu/api/res/1.2/7LKu1VqFsBWR.ZGGf.U.zQ--/YXBwaWQ9eXRhY2h5b247aD04Njt3PTEzMDs-/https://s.yimg.com/os/creatr-images/2019-11/fae1bf00-0581-11ea-93ff-f43ed8c16284"
+    ]
+  }
+]
```
HenadziStantchik: Cache files should not be checked in

HenadziStantchik:
tar.gz files should not be checked in to GIT

final_task/rss_reader/ConsoleParse.py
```diff
@@ -0,0 +1,14 @@
+import argparse
+
+
+def get_arguments_from_console():
+    arg_parser = argparse.ArgumentParser(description="Pure Python command-line RSS reader.")
+    group = arg_parser.add_mutually_exclusive_group()
```
HenadziStantchik: (Optional) It is better to make that if the source is specified, then application looks through local cache and prints only news from the specified source

final_task/rss_reader/Handler.py
```diff
@@ -0,0 +1,111 @@
+import json
+import logging
+from itertools import islice
+from json import JSONDecodeError
```
HenadziStantchik: you have already imported `json` lib above. you can access it using `json.JSONDecodeError`

final_task/rss_reader/Handler.py
```diff
@@ -0,0 +1,111 @@
...
+
+        if not [entry for entry in entries if entry["Title"] == entry_dict["Title"]]:
+            entries.append(entry_dict)
+
+        with open("cache.json", "w") as cache:
+            json.dump(entries, cache, indent=2)
+
+    @logging_decorator
+    # read entries from cache.json into list daily_news that have date that is equal to --date DATE
+    # and then raise an exception or print to console
```
HenadziStantchik: Better move this to docstring

HenadziStantchik:
Did not discover any issues with Iteration 3 work, good job. Resolve small issues in comments and proceed to Iteration 4

final_task/README.md
```diff
@@ -1,3 +1,122 @@
...
+usage: rss_reader.py [-h] [--version] [--json] [--verbose] [--limit LIMIT]
+                     [--date DATE]
+                     [source]
+rss_reader.py: error: argument --date: not allowed with argument source
+```
+Argument --date work with --json, --limit and --verbose arguments
+```
+$ python rss_reader.py --date 20191113 --json --verbose
+```
+News is stored in local file cache.json as list of json objects
```
HenadziStantchik: Please add info about iteration 4

final_task/rss_reader/RSSReaderException.py
```diff
@@ -0,0 +1,3 @@
+class RSSReaderException(Exception):
+    def __init__(self, message):
+        self.message = message
```
HenadziStantchik: You do not really need this logic here. With `pass` it will do the same

HenadziStantchik:
When there is no `cache.json` file, `FileNotFoundError` is thrown to the stdout with traceback.

Please keep in mind that you need to convert news to at least 2 different formats.

HenadziStantchik:
1.Conversion does not work with `--date`

If you are not going to implement search in cache by date and by rss url, then it is better to make that if `--date` is specified, RSS url is just ignored

torch

2. Even it there is internet connection, there are no images in the converted html file
Also there is encoding problem there:
torch

final_task/rss_reader/Handler.py
```diff
@@ -0,0 +1,272 @@
...
+    def write_entries_to_html(self, path: str, entries=[]) -> None:
+        # in case of reading news from cache list of entries are got as dict
+        # and in case of online reading news only the path is passed to the method without list of entries
+        if not entries:
+            entries = self.entries
+
+        if path[-1] == '\\':
+            path = path[:-1]
+        if path[-1] != '/':
+            path += '/'
```
HenadziStantchik: It is better to use `os.path` lib for any operations with filesystem paths

HenadziStantchik:
It is better to have more informative names for converted files.

final_task/rss_reader/Handler.py
```diff
@@ -0,0 +1,272 @@
...
+    def write_to_pdf(self, path: str, entries=[]) -> None:
+        # in case of reading news from cache list of entries are got as dict
+        # and in case of online reading news only the path is passed to the method without list of entries
+        if not entries:
+            entries = self.entries
+
+        if path[-1] == '\\':
+            path = path[:-1]
+        if path[-1] != '/':
+            path += '/'
```
HenadziStantchik: Same: `os.path`

IlyaTorch:
I have no problems with encoding in all my browsers. I rewrote write_entries_to_html methon but I'm not sure that it'll help. What can I do with it?

final_task/rss_reader/Handler.py
```diff
@@ -0,0 +1,260 @@
...
+sys.path.append(this_dir)
+import feedparser
+from Entry import Entry
+from Logging import logging_decorator
+from RSSReaderException import RSSReaderException
+from RSS_PDF import PDF
+import urllib.request
+
+import dominate
+from dominate.tags import *
```
dzhigailo: Please, do not use `import *`

final_task/rss_reader/Handler.py
```diff
@@ -0,0 +1,260 @@
...
+
+    @logging_decorator
+    def correct_title(self, title: str) -> str:
+        return title.replace('"', "_").replace("?", "_").replace(":", "_").replace("'", "_")[:30].replace(" ", "_")
+
+    @logging_decorator
+    def write_cache(self, entry_dict: dict):
+        """write new news to cache in json format"""
+        try:
+            entries = json.load(open("cache.json"))
```
dzhigailo: Please, use context manager for working with files. You can check if file exists with `os.path.exists()`

final_task/rss_reader/Handler.py
```diff
@@ -0,0 +1,260 @@
...
+        # in case of reading news from cache list of entries are got as dict
+        # and in case of online reading news only the path is passed to the method without list of entries
+        if os.path.isdir(path) is False:
+            raise RSSReaderException("Error. It isn't a folder")
+        path = os.path.join(path, "RSS_News.html")
+
+        if not entries:
+            entries = self.entries
+
+        if type(entries[0]) == dict:
```
dzhigailo: Better use `isinstance()` function

final_task/rss_reader/Handler.py
```diff
@@ -0,0 +1,260 @@
...
+            entries = self.entries
+
+        pdf = PDF()
+        pdf.add_font('DejaVuSans', '', 'DejaVuSans.ttf', uni=True)
+        pdf.set_font("DejaVuSans")
+        pdf.alias_nb_pages()
+        pdf.add_page()
+
+        # in case of reading news from cache, entries is the list of dicts and they are converted to Entry-object
+        if type(entries[0]) == dict:
```
dzhigailo: Same as above

final_task/rss_reader/Handler.py
```diff
@@ -0,0 +1,260 @@
...
+                while text.count('['):
+                    text = text[:text.find('[')] + text[text.find(']') + 1:]
+                if text[0] == ' ':
+                    text = text[1:]
+
+            pdf.set_font_size(24)
+            pdf.write(10, entry.title + '\n\n')
+            pdf.set_font_size(14)
+            pdf.write(10, f"Feed: {entry.feed}\n")
+            pdf.set_font_size(14)
```
dzhigailo, line 335: Why do you duplicate line 245? Haven't you already set font size 14?

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,57 @@
...
+    if arg_parser_args.verbose:
+        logging.getLogger().setLevel(logging.INFO)
+    if arg_parser_args.limit <= 0 and arg_parser_args.date is False:
+        raise RSSReaderException("Error. Incorrect value of argument limit")
+
+    if arg_parser_args.version:
+        handler.option_version()
+    elif arg_parser_args.date:
+        handler.option_date(str(arg_parser_args.date), arg_parser_args.json,
+                            arg_parser_args.to_html, arg_parser_args.to_html,
```
dzhigailo: Why do you pass same arguments twice?

final_task/rss_reader/Handler.py
```diff
@@ -0,0 +1,260 @@
...
+        except FileNotFoundError:
+            os.makedirs("images")
+            urllib.request.urlretrieve(img_url, f"images/{img_name}.jpg")
+
+    @logging_decorator
+    def option_date(self, date: str, do_json: bool, do_html: bool, html_path: str, do_pdf: bool, pdf_path: str):
+        """add entries from cache.json into daily_news: list if they have date that is equal to user's --date DATE
+            and then raise an exception or print to console or to outputs to html"""
+        try:
+            entries = json.load(open("cache.json"))
```
dzhigailo: Same as above

final_task/rss_reader/Handler.py
```diff
@@ -0,0 +1,260 @@
...
+            entries = []
+        except FileNotFoundError:
+            raise RSSReaderException("Error. You have no cache. Try to run app with internet-connection")
+
+        # list of entries with the same date as user's --date DATE
+        daily_news = [entry for entry in entries if entry["DateInt"] == date]
+        if not daily_news:
+            raise RSSReaderException("Error. News aren't found")
+        # different cases of command line arguments
+        elif do_pdf and do_html:
```
dzhigailo: You can optimize this statements. Instead of elif, you could just check your arguments one by one with if and do some logic, instead of duplicating your code.

HenadziStantchik:
> I have no problems with encoding in all my browsers. I rewrote write_entries_to_html methon but I'm not sure that it'll help. What can I do with it?

Then this means that the problem on my side. It is ok then.

HenadziStantchik:
I did not discover any more issues with iteration 4

IlyaTorch:
Please look at my tests. Can you explain me why, when I add new tests, % of coverage doesn't change?

IlyaTorch:
And how do the files for which I didn't write a single test show 100% coverage? I check coverage so
`coverage run rss_reader.py`
`coverage report -m` or  `coverage html`

IlyaTorch:
image
I wrote tests only for files Entry.py and Hanler.py

HenadziStantchik:
Coverage for exceptions is 100% because they do not contain any statements that need to be tested, only class declarations.

Coverage for logging is 100% because you decorated almost every function with logging decorator, so it implicitly called in each your test, so the coverage thinks that you tested it.

If you wrote tests only for entry and handler and there is some percentage on other modules, that means that the functions you wrote tests for call some functions from other modules and this makes coverage think that you tested these functions too.

Try to check some documentation on coverage, I am sure you will find solution to your problem

final_task/README.md
```diff
@@ -1,3 +1,144 @@
...
+python rss_reader.py "https://timesofindia.indiatimes.com/rssfeedstopstories.cms" --json --limit 1
+```
+
+
+5 mains files of project:
+* rss_reader.py - the file which runs the application
+* ConsoleParse.py - contains code which parses arguments from console
+* Entry.py - contains class Entry which represent an article
+* Handler.py - contains class Handler which performes functions of processing objects Entry
+* Logging.py - contains decorator for printing loggs in stdout
```
HenadziStantchik, line 14: It would be nice if you kept it up to date.

HenadziStantchik:
Iteration 1:
`--json` does not work with Cyrillic symbols.
Iteration 2:
Installs fine, all iterations work.
Iteration 3:
User should be able specify date in `YYYYMMDD`:
image
Iteration 4:
Everything works fine.


Commit messages are very informative and correspond to the recommended commit message style.
Code is well structured and readable, but there are some large methods which lower readability a bit. The code corresponds to the PEP8 guidelines.
Tests are decent, but some of them are failing `test_write_entries_to_html` (and without). keep in mind that tests should not depend on any external factors such as OS or etc.  Coverage percentage is high enough.

PR #11, opened 2019-11-06
Leonid_Shutov_14m64y87w@gmail.com - PLEASE REVIEW

leonid-shutov:
Iteration 1


README.md
```diff
@@ -18,9 +18,15 @@ Congrats! You have successfully forked our repository.
 ## Additional project structure requirements
 
 1. `setup.py` file for setuptools  *must*  be in the root of  `final_task`  folder. Use  `setup.py`  that is already there. (that means path to this file must end with  `final_task/setup.py` )  
+<<<<<<< HEAD
```
dzhigailo: Please, resolve conflicts

final_task/rss_reader/requirements.txt
```diff
@@ -0,0 +1,7 @@
+import socket
```
dzhigailo: This file should contain extarnal libs, that are required by the application. Import are not needed here.
You can read about this files here: https://pip.pypa.io/en/stable/user_guide/#id12

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,114 @@
...
+from bs4 import BeautifulSoup
+
+
+log = logging.getLogger("Log")
+
+
+def find_news(url):
+    """Parses news"""
+    logging.info("Searching for news on " + url)
+    while True:
```
dzhigailo: What is the reason to use `while` in here? There are no iterations.

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,114 @@
...
+        try:
+            resp = requests.get(url)
+            soup = BeautifulSoup(resp.content, features="xml")
+            items = soup.findAll('item')
+            log.info(str(len(items))+" pieces of news have been found")
+        except requests.exceptions.MissingSchema:
+            log.error("Wrong URL")
+            print("Wrong URL: " + url)
+            print("Try to input a correct URL: ")
+            url = input()
```
dzhigailo: In case of incorrect URL, application should be gracefully shutdown. Also, please dont use `input` function. The only input that application will have is initial arguments.

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,114 @@
...
+def collect_news(items, limit) -> list:
+    """Sorts the components of the news into dictionaries and adds them into a list"""
+    news = []
+    for item in items:
+        try:
+            news_item = {}
+            news_item['title'] = item.title.text
+            news_item['description'] = item.description.text
+            news_item['link'] = item.link.text
+            if item.thumbnail != None:
```
dzhigailo: Here and below: 
In accordance with PEP8, this check should look like `item.thumbnail is None`.

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,114 @@
...
+        print("Title: " + item['title'])
+        print("Description: " + item['description'])
+        print("Image: " + item['image'])
+        print("Link: " + item['link'])
+        print('------------------')
+
+
+def json_convert(news):
+    """Converts the news into JSON format"""
+    with open("README.md", "w") as write_file:
```
dzhigailo: Why do you overwrite the README.md file?

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,114 @@
...
+        json.dump(news, write_file)
+
+
+if __name__ == '__main__':
+    
+    while True:
```
HenadziStantchik: You do not need iterations here. Application is supposed to be 'one-shot' as stated in the task. Meaning:
We run this application from the command line like this:

`python rss_reader.py <RSS_URL>`

It fetches news for us and then shuts down.

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,125 @@
+import socket
+import sys
+import requests
+import argparse
+import json
+import logging
+import pycodestyle
```
HenadziStantchik: pycodestyle is a tool so you do not actually need to import it anywhere. If you want to use it just type this in the command line: ```pycodestyle <file_name>.py``` and it will analyse it.

The same for requirements.txt file.

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,125 @@
...
+        parser = argparse.ArgumentParser(description="RSS")
+        parser.add_argument('source', type=str, help='RSS URL')
+        parser.add_argument('--version', action='store_true', help='Print version info')
+        parser.add_argument('--json', action='store_true', help='Print result as JSON in stdout')
+        parser.add_argument('--verbose', action='store_true', help='Outputs verbose status messages')
+        parser.add_argument('--limit', type=int, default=999, help='Limit news topics if this parameter provided')
+        args = parser.parse_args()
+    except SyntaxError:
+        print("Wrong syntax in args")
+        sys.exit()
```
HenadziStantchik: It is hightly unlikely to get `SyntaxError` here, so it is better to remove it

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,125 @@
...
+if __name__ == '__main__':
+    
+    try:
+        parser = argparse.ArgumentParser(description="RSS")
+        parser.add_argument('source', type=str, help='RSS URL')
+        parser.add_argument('--version', action='store_true', help='Print version info')
+        parser.add_argument('--json', action='store_true', help='Print result as JSON in stdout')
+        parser.add_argument('--verbose', action='store_true', help='Outputs verbose status messages')
+        parser.add_argument('--limit', type=int, default=999, help='Limit news topics if this parameter provided')
+        args = parser.parse_args()
```
HenadziStantchik: it is better to move it to separate function

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,125 @@
...
+    if args.version:
+        print("Version is 1.1")
+    if args.verbose:
+        logging.basicConfig(level=logging.INFO)
+    
+    if args.limit:
+        if args.limit > 0:
+            limit = args.limit
+        else:
+            limit = 5
```
HenadziStantchik: If `--limit` is incorrect it is better to just shutdown the program.

HenadziStantchik:
1.User should be able to use your application like this:
```python rss_reader.py --version```. Right now it is not possible to get your app version without specifying source url.

2(optional). You should add a bit more logging to your app, to make `--verbose` more explicit and noticeable.

3. ~~`--json` does not work~~
Update: Check my comment about your `json_convert` function.

File: final_task/rss_reader/rss_reader.py [hunk 12]
```diff
@@ -0,0 +1,125 @@
+import socket
```
HenadziStantchik: unused import here

README.md
```diff
@@ -18,9 +18,15 @@ Congrats! You have successfully forked our repository.
...
 2. Entry point to your application, aka its main module  *must*  be named as  `rss_reader.py` . Use  `rss_reader.py`  that is already in  `rss_reader`  folder.  
 3. You should describe how does your project work, how to launch it and etc in README.md in the  `final_task/README.md`  file.  
 4. If you used any non-standart libraries they must be listed in  `rss_reader/requirements.txt`  file.  
+=======
+2. Entry point to your application, aka its main module  *must*  be named as  `rss_parser.py` . Use  `rss_parser.py`  that is already in  `rss_parser`  folder.  
+3. You should describe how does your project work, how to launch it and etc in README.md in the  `final_task/README.md`  file.  
+4. If you used any non-standart libraries they must be listed in  `rss_parser/requirements.txt`  file.  
+>>>>>>> Add instructions in README.Add basic .py modules
 5. All unit test files should be in separate folder called `tests`.
```
HenadziStantchik: To solve conflicts just remove lines marked as green in your README.md

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,125 @@
...
+            limit = args.limit
+        else:
+            limit = 5
+    channel = find_channel(args.source)
+    items = find_news(args.source)
+    news = collect_news(items, limit)
+    print_news(news, channel)
+
+if args.json:
+    json_convert(news)
```
HenadziStantchik: You probalby fogot to add indentation here

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,125 @@
...
+        if item['alt']:
+            print("Alt: " + item['alt'])
+        print("Link: " + item['link'])
+        print('------------------')
+
+
+def json_convert(news):
+    """Converts the news into JSON format"""
+    with open("JSON.json", "w") as write_file:
+        json.dump(news, write_file)
```
HenadziStantchik: `--json` option shiould just print parsed news to console, but in json format. No need actually write it to file.

final_task/rss_reader/rss_reader.py
Leonid-Shutov: I see, but it works for me without any indentation(for some reason)

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,130 @@
...
+    logging.info("Searching for news on " + url)
+    try:
+        resp = requests.get(url)
+        soup = BeautifulSoup(resp.content, features="xml")
+        items = soup.findAll('item')
+        log.info(str(len(items)) + " pieces of news have been found")
+    except requests.exceptions.MissingSchema:
+        log.error("Wrong URL")
+        print("Wrong URL: " + url)
+        sys.exit()
```
HenadziStantchik: Here and below: It is better to have only one exit point from the application. For example you can raise your custom exception here and handle it in the main function

final_task/rss_reader/requirements.txt
```diff
@@ -0,0 +1,4 @@
+beautifulsoup4==4.8.1
+lxml==4.4.1
+requests==2.22.0
+soupsieve==1.9.4
```
HenadziStantchik: You do not use this lib anywhere. Do you really need it?

HenadziStantchik:
Please resolve older issues and remove unused files from `rss_parser` folder.

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,174 @@
...
+from bs4 import BeautifulSoup
+
+
+log = logging.getLogger("Log")
+
+
+class CustomEx(Exception):
+    def __init__(self, msg):
+        self.msg = msg
+        super().__init__(msg)
```
HenadziStantchik: It is better to name if after the situation you are going to raise it, for example `ConnectionError` and etc.
Also it is better to have separate module for it.
And also you can just implement it like this:
```
class CustomEx(Exception):
    pass
```

That will work practically the same way you implemented, but without redundant code.

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+def is_connected() -> bool:
+    log.info("Checking internet connection...")
+    try:
+        socket.create_connection(("www.google.com", 80))
+        return True
+    except OSError:
+        pass
+    raise CustomEx("No internet connection")
```
HenadziStantchik: Try to rethink your sequence of actions here, this code can be simpler.
1. try to create connection
2. if anything is wrong - raise an exception
3. if everything ok - return true

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,174 @@
...
+    """Parses news"""
+    logging.info("Searching for news on " + url)
+    try:
+        resp = requests.get(url)
+        soup = BeautifulSoup(resp.content, features="xml")
+        items = soup.findAll('item')
+        log.info(str(len(items)) + " pieces of news have been found")
+    except requests.exceptions.MissingSchema:
+        log.error("Wrong URL")
+        raise CustomEx("Wrong URL: " + url)
```
HenadziStantchik: Here, above and below: It is better to have different custom exceptions for different situations.

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,174 @@
...
+                news_item['alt'] = alts
+        except TypeError:
+            log.error("Could not load an image")
+            news_item['image'] = "No image"
+        except AttributeError:
+            log.error("Probably no description")
+            news_item['description'] = "No description"
+        news.append(news_item)
+        if len(news) == limit:
+            log.info("The limit of " + str(limit) + "pieces of news reached")
```
HenadziStantchik: It is better to use formatted string here.

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,174 @@
...
+                    alts.append(alt['alt'])
+            if item.find('media:text'):
+                log.info('media:text')
+                soup = BeautifulSoup(item.find('media:text').text, 'lxml')
+                for alt in soup:
+                    alts.append(alt.find('img')['alt'])
+            # if len(item.credit.text) > 0:
+            #     log.info(item.credit.text)
+            #     alts.append(item.credit.text)
+            if len(alts) == 0:
```
HenadziStantchik: here and above: You can check if the sequence is empty or not simply by using `if `alts`:`, no need to check its length

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,174 @@
...
+                channel = find_channel(args.source)
+                items = find_news(args.source)
+                news = collect_news(items, limit)
+                print_news(news, channel)
+                if args.json:
+                    json_convert(news)
+    except CustomEx as ex:
+        print(ex.msg)
+        print('Exiting the program')
+        log.info('Exiting the program')
```
HenadziStantchik: It is better move this to `finally` block.

final_task/setup.py
```diff
@@ -0,0 +1,25 @@
...
+      name='rss_reader',
+      version='2.0',
+      packages=find_packages(),
+      description='RSS-reader',
+      long_description=long_description,
+      long_description_content_type='text/markdown',
+      url='https://github.com/Leonid-Shutov/FinalTaskRssParser',
+      author='Leonid Shutov',
+      author_email='14m64y87w@gmail.com',
+      install_requires=['beautifulsoup4==4.8.1','lxml==4.4.1','requests==2.22.0','soupsieve==1.9.4'],
```
HenadziStantchik: this line and your requirements.txt do not correspond each other.

HenadziStantchik:

1. Currently you print 2 alts for a single image. Also if there is no image it is the same as title.
2. (optional) it would be nice to separate news visually from each other a little bit. Right now it is a bit hard to tell where ends one and starts the other.
3. If user specifies `--json` then news should be printed in json format instead of usual feed, not after it.

final_task/rss_reader/rss_reader.py
leonid-shutov: Can I have a single module for all custom exceptions?

HenadziStantchik:
You forgot to push your `CustomException` module

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,267 @@
...
+import argparse
+import json
+import logging
+import socket
+import bs4
+from bs4 import BeautifulSoup
+import CustomException
+
+
+log = logging.getLogger("Log")
```
HenadziStantchik: It is better to avoid using global variables, but if you cannot help it, then they should be declared in uppercase, e.g. `VERSOIN`

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,267 @@
...
+
+
+def is_connected() -> bool:
+    log.info("Checking internet connection...")
+    try:
+        socket.create_connection(("www.google.com", 80))
+    except OSError:
+        pass
```
HenadziStantchik: Do you really need this block if you trying to catch a broader exception a bit lower?

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,267 @@
...
+        "11": 'Nov',
+        "12": 'Dec',
+    }
+    date = date[6:] + ' ' + calendar[date[4:6]] + ' ' + date[0:4]
+    try:
+        file = open('cache.json')
+    except FileNotFoundError:
+        print('No news in cache yet')
+    else:
+        with file:
```
HenadziStantchik: It is better to use context manager this way:
```
try: 
    with open(...) as file:
        ......
except FileNotFoundError:
    ....
```

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,267 @@
...
+        "05": 'May',
+        "06": 'Jun',
+        "07": 'Jul',
+        "08": 'Aug',
+        "09": 'Sep',
+        "10": 'Oct',
+        "11": 'Nov',
+        "12": 'Dec',
+    }
+    date = date[6:] + ' ' + calendar[date[4:6]] + ' ' + date[0:4]
```
HenadziStantchik: Please check `datetime` library instead of concatenating date this way

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,267 @@
...
+        file = open('cache.json')
+    except FileNotFoundError:
+        print('No news in cache yet')
+    else:
+        with file:
+            json_list = json.load(file)
+            print(date + ':\n')
+            if source:
+                source = source[8:source.find('/', 8)]
+                for item in json_list:
```
HenadziStantchik: You can use `enumerate` here and check `if index > limit` instead of lowering `limit` by 1 each iteration

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,267 @@
...
+                    if date in item['date'] and source in item['link']:
+                        flag = True
+                        print_one(item)
+                        limit -= 1
+                        if not limit:
+                            break
+                else:
+                    for item in json_list:
```
HenadziStantchik: Same as above

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,267 @@
...
+                print('Version is 1.2')
+            if args.verbose:
+                logging.basicConfig(level=logging.INFO)
+
+            if args.limit:
+                limit = args.limit
+                log.info(f'Limit = {limit}')
+            if args.source:
+                if 'www' in args.source:
+                    args.source = args.source[:8] + args.source[12:]
```
HenadziStantchik: A question:

Why do you need to do this?

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,267 @@
...
+                else:
+                    print_news(news, channel)
+                write_cache(news)
+                if args.date:
+                    print('\n\n\n\n')
+    except CustomException.ConnectionError:
+        print('Can not connect to the internet. Check your connection')
+    # except NameError:
+    #     log.error('Wrong limit')
+    #     print('Wrong limit')
```
HenadziStantchik: Please avoid using commented-out code in the pushed commits

final_task/rss_reader/rss_reader.py
leonid-shutov: Because a user can type both https://www.lenta.ru/rss and https://lenta.ru/rss
And I need a single variant of url to compare it with other links when I use source + date options
So I remove "www." if I find it in url

final_task/rss_reader/converter.py
```diff
@@ -0,0 +1,120 @@
...
+                count += 1
+                position = count_on_page * 52 + 12
+                try:
+                    response, content = h.request(image)
+                    out = open('image' + str(count) + '.jpg', 'wb')
+                    out.write(content)
+                    out.close()
+                    pdf.image('image' + str(count) + '.jpg', 130, position, 70, 50)
+                except Exception:
+                    pass
```
HenadziStantchik: It is better to use `slices` instead of having counter and comparing it with `limit` each iteration here.

final_task/rss_reader/converter.py
```diff
@@ -0,0 +1,120 @@
...
+        if len(string) > 3000 and count_on_page < 1:
+            pdf.set_font("FreeSans", size=6)
+            pdf.multi_cell(w=190, h=4, txt=string)
+        else:
+            pdf.set_font("FreeSans")
+            pdf.multi_cell(w=100, h=4, txt=string, align='L')
+    if not os.path.exists(path):
+        os.path.join(path)
+    time_name = datetime.strftime(datetime.now(), "%d-%m-%y-%H-%M")
+    filename = path + '/' + time_name + '.pdf'
```
HenadziStantchik: Please try to rethink/split logic here. Right now there are too many repetitive lines. Probably try to apply your custom context manager/decorator and etc.

final_task/rss_reader/converter.py
```diff
@@ -0,0 +1,120 @@
...
+        except FileNotFoundError:
+            pass
+
+
+def html_convert(news, limit, path):
+    doc = dominate.document(title='News')
+    for ind, item in enumerate(news):
+        if ind == limit:
+            break
+        with doc:
```
HenadziStantchik: It is better to use `slices` here, instead of comparing index with limit on each iteration

final_task/rss_reader/converter.py
```diff
@@ -0,0 +1,120 @@
...
+                    else:
+                        p('Images:')
+                        imgstr = ''
+                        for image in item['image']:
+                            imgstr += image + '\n'
+                        p(imgstr)
+    if not os.path.exists(path):
+        os.path.join(path)
+    time_name = datetime.strftime(datetime.now(), "%d-%m-%y-%H-%M")
+    filename = path + '/' + time_name + '.html'
```
HenadziStantchik: It is better to use `os.path` wherever you are working with filesystem paths. This line would not work correctly on Windows

final_task/rss_reader/rss_reader.py
```diff
@@ -176,92 +181,108 @@ def write_cache(news):
...
-                for item in json_list:
+                for ind, item in enumerate(json_list):
                     if date in item['date'] and source in item['link']:
                         flag = True
-                        print_one(item)
-                        limit -= 1
-                        if not limit:
+                        news.append(item)
+                        if len(news) > limit - 1:
+                            break
```
HenadziStantchik, line 219: Same here: `slices`

final_task/rss_reader/rss_reader.py
HenadziStantchik: It is better to just shutdown the program if the user specified incorrect link or the link that does not return rss feed.
There may be rss feed links that may contain `www` naturally

HenadziStantchik:
1. There is encoding problem in converted HTML news. (Firefox browser)
2. Whenever user converts news to html or pdf from local cache, many empty lines are printed in stdout.

final_task/rss_reader/converter.py
```diff
@@ -18,15 +18,30 @@ def is_connected() -> bool:
     return True
 
 
-def pdf_convert(news, limit, path):
+def font_decorator(function_to_decorate):
+    def wrapper(text, page, font):
```
HenadziStantchik, line 25: A docstring here would be nice

final_task/rss_reader/converter.py
```diff
@@ -18,15 +18,30 @@ def is_connected() -> bool:
...
+        function_to_decorate(text, page, font)
+        if font == 'Arial':
+            page.cell(1, 1, text + '\n')
+        else:
+            page.multi_cell(w=100, h=6, txt=text, align='L')
+    return wrapper
+
+
+@font_decorator
+def font_print(text, page, font):
```
HenadziStantchik, line 35: Docstring here would be nice

final_task/rss_reader/converter.py
```diff
@@ -0,0 +1,135 @@
...
+        if not is_connected():
+            text = '\n'
+            font_print('Images', pdf, 'Arial')
+            for image in item['image']:
+                text += image + '\n'
+            text += '\n'
+            pdf.set_font("FreeSans", size=8)
+            pdf.multi_cell(w=100, h=4, txt=text, align='L')
+        font_print('Description', pdf, 'Arial')
+        text = '\n' + item['description']
```
HenadziStantchik: This chunk of code is still a bit hard to read, try to think of some more ways to simplify it

HenadziStantchik:
1. A bit more different colours would be nice (optional)

2. `Alt` field is duplicated in JSON:

final_task/README.md
```diff
@@ -1,3 +1,21 @@
...
+"alt": "Lebanese banks urge calm amid financial crisis, protests"
+}
+
+Convertion:
+Dominate lib is used to create html files.
+FPDF lib is used to create pdf files.
+When loading images for pdf, caching is used to fasten loading(httplib2)
+To let  cyrillic be encoded in the right way to see it in pdf, please open FreeSans.ttf
+
+When colorizing output, application uses cprint from termcolor library
```
HenadziStantchik, line 21: IT would be nice if you could make your documentation a bit larger and specific: how to install your app, what args does it have, how do you store news and etc.

final_task/rss_reader/converter.py
```diff
@@ -0,0 +1,152 @@
...
+    Sets a font on a page
+    Has a text argument to have a possibility to be wrapped
+    """
+    if font == 'Arial':
+        page.set_font(font, 'B', size=12)
+    else:
+        page.set_font(font)
+
+
+def pdf_convert(news, limit, path, color):
```
HenadziStantchik, line 46: This function is really large, it is better to split it somehow into the smaller ones.

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,294 @@
...
+    except requests.exceptions.ConnectionError:
+        raise CustomException.WrongUrl
+    except requests.exceptions.InvalidURL:
+        raise CustomException.WrongUrl
+    except requests.exceptions.MissingSchema:
+        raise CustomException.WrongUrl
+    return channel
+
+
+def collect_news(items, limit) -> list:
```
HenadziStantchik, line 55: Same as above, this function is large which lowers code readability. It is better to somehow split it.

HenadziStantchik:
Iteration 1:
If rss url specified along with `--version` news feed still should not be printed.
Iteration 2:
Installs, does not work:
image
Iteration 3:
Everything is fine. Good job implementing search by rss source.
Iteration 4:
Everything works fine.
Iteration 5:
Everything works fine, looks good.

Tests are almost nonexistent, coverage percentage is not enough.
Code is mostly readable, although there are some large functions which lower readability. 
Commit messages are informative, but do not correspond to the recommended commit message style.

PR #12, opened 2019-11-07
Anton_Zimahorau_antonzimahorau@gmail.com

AntonZimahorau:
First iteration


final_task/rss_reader/classes/arguments.py
```diff
@@ -0,0 +1,36 @@
...
+                            provided""", type=int)
+        parser.add_argument("source", help="RSS URL")
+        # Parse arguments
+        args = parser.parse_args()
+        return args
+
+    def __init__(self):
+        """Store all arguments as class attributes"""
+
+        args = self.get_arguments()
```
HenadziStantchik: You actually already have object with all arguments parsed here, so you do not need separate class for this.

final_task/rss_reader/classes/news_class.py
```diff
@@ -0,0 +1,32 @@
...
+    Uses for storing title, date, link, text and links from text
+    as an attributes of class.
+
+    Methods
+    -------
+    print_feed_title(self) - print to stdout title of rss feed.
+
+    print_news(self) - print to stdout title, date, link, text
+        and links from text in particular sequence.
+    """
```
HenadziStantchik: If you have docstrings in corresponding methods it is better to not have these descriptions in class docstring.
But you can move this to your documentation

final_task/rss_reader/functions/check_func.py
```diff
@@ -0,0 +1,85 @@
...
+--------------------------------------------
+check_verbose(command_line_args): -> logger
+    If vebose argument is True, set logger to info level
+--------------------------------------------
+check_feed_status(feed, logger)
+    Check HTTP status code
+--------------------------------------------
+def check_news_collection(news_collection, logger) -> news_collection
+    Check if news_collection is not empty
+"""
```
HenadziStantchik: Same as above, it is better to move this to documentation. You do not need this here if you have docstrings in corresponding functions.

final_task/rss_reader/functions/check_func.py
```diff
@@ -0,0 +1,85 @@
...
+import requests
+import logging
+import sys
+
+
+def check_internet_connection(logger):
+    """Check if network connection is avaliable"""
+    try:
+        logger.info("Checking internet connection...")
+        requests.get('http://216.58.192.142', timeout=1)
```
HenadziStantchik: A question:

Why this IP specifically? What if this IP is unavailable while you have internet connection?

final_task/rss_reader/functions/check_func.py
```diff
@@ -0,0 +1,85 @@
...
+        return True
+    except requests.ConnectionError:
+        logger.warning("Check internet connection")
+        # Ask user is he want to proceed
+        answer = input("Do you want to proceed? (Y/n): ")
+        if answer.lower() == 'y':
+            check_internet_connection(logger)
+        else:
+            exit()
```
HenadziStantchik: Here and below: it is better to have only one exit point from the application.
For example you should raise your custom exception here and handle it in the main function.

final_task/rss_reader/functions/check_func.py
```diff
@@ -0,0 +1,85 @@
...
+    else:
+        logger.setLevel(level=logging.WARNING)
+        return logger
+
+
+def check_feed_status(feed, logger):
+    """Check HTTP status code"""
+
+    if feed.status:
+        if feed.status == 404:
```
HenadziStantchik: It is better to check not just 404 code, but all codes starting with 4** or 5**. Check feedparser methods, probably there is method to check response status. If there is not, then you can alvays check status code using `><`

final_task/rss_reader/functions/print_func.py
```diff
@@ -0,0 +1,72 @@
...
+    """
+
+    if command_line_args.limit:
+        limit = command_line_args.limit
+        if len(news_collection) > limit >= 0:
+            return news_collection[:limit]
+        elif limit < 0:
+            logger.warning("Limit argument can't be a negative")
+            limit = int(input("Please, enter the valid value: "))
+            command_line_args.limit = limit
```
HenadziStantchik: It is better to have this check before everything else in this function. Moreover, if user entered wrong limit it is his fault, so just raise an exception here and stop the script.

All user inputs should be always validated and having `input()` anywhere is unsafe.

final_task/rss_reader/functions/process_func.py
```diff
@@ -0,0 +1,117 @@
...
+    Get feed title
+    Create a list of news instance objects
+"""
+
+
+import feedparser
+from lxml import etree
+from classes.news_class import News
+import functions.check_func as ch_f
+
```
HenadziStantchik: Here and below: It is better to use `from <module> import <class/function>` in such situations.

final_task/rss_reader/functions/process_func.py
```diff
@@ -0,0 +1,117 @@
...
+        text = extract_text_from_description(root)
+        links = extract_links_from_description(root)
+
+        # Create News object and define attributes
+        news = News()
+        news.title = entry.get("title", "").replace('&#39;', "'")
+        news.date = entry.get("published", "")
+        news.link = entry.get("link", "")
+        news.text = text
+        news.links = links
```
HenadziStantchik: This logic should be inside an `__init__` method of the `News` class. Also it is better to make it a `dataclass`.

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,54 @@
...
+"""
+
+
+def main():
+    """Main function that runs the program"""
+
+    from classes.arguments import ComLineArgParser
+    import functions.check_func as ch_f
+    import functions.process_func as proc_f
+    import functions.print_func as print_f
```
HenadziStantchik: It is better to have imports at the top of the module, not inside functions/classes

final_task/rss_reader/functions/check_func.py
```diff
@@ -0,0 +1,85 @@
...
+import sys
+
+
+def check_internet_connection(logger):
+    """Check if network connection is avaliable"""
+    try:
+        logger.info("Checking internet connection...")
+        requests.get('http://216.58.192.142', timeout=1)
+        return True
+    except requests.ConnectionError:
```
HenadziStantchik, line 15: It is better to make application raise an exception and gracefully shutdown if there is no internet connection.

HenadziStantchik:

1. If there is image on the news page that has src="", then there is empty line in "Links":

image

2. For `--verbose` option it would be nice to have logging while printing news/creating news entities too.

3. There is a small bug with your links list in `--json` argument:

image

final_task/rss_reader/classes/arguments.py
AntonZimahorau: Fixed

final_task/rss_reader/classes/news_class.py
AntonZimahorau: Fixed

final_task/rss_reader/functions/check_func.py
AntonZimahorau: Fixed

final_task/rss_reader/functions/check_func.py
AntonZimahorau: Fixed

final_task/rss_reader/functions/check_func.py
AntonZimahorau: Fixed

final_task/rss_reader/functions/check_func.py
AntonZimahorau: Fixed

final_task/rss_reader/functions/print_func.py
AntonZimahorau: Fixed

final_task/rss_reader/functions/process_func.py
AntonZimahorau: Fixed

final_task/rss_reader/functions/process_func.py
AntonZimahorau: Fixed

final_task/rss_reader/rss_reader.py
AntonZimahorau: Fixed. 
I used import inside function, because i wanted to handle all exception in one try/except block.

final_task/rss_reader/functions/check_func.py
AntonZimahorau: Fixed

final_task/rss_reader/functions/check_func.py
```diff
@@ -0,0 +1,52 @@
...
+        logger.setLevel(level=logging.WARNING)
+        return logger
+
+
+def check_feed_status(feed):
+    """Check HTTP status code"""
+
+    try:
+        status = feed.status
+        if 600 > status >= 400:
```
dzhigailo: There is a better way to check if it's 500 or 400 HTTP response. Please, rework this solution

final_task/rss_reader/functions/check_func.py
```diff
@@ -0,0 +1,52 @@
...
+            raise exc.GettingFeedError("Server Error! Feed wasn't received.")
```
dzhigailo: 400 code is not Server Error.

final_task/README.md
```diff
@@ -1,3 +1,95 @@
...
+       "link": "News link",
+       "text": "Text content of news",
+       "links": ["links that presented in text",]
+     },
+     {
+        ...
+     }
+	    ]
+}
+```
```
HenadziStantchik: Please add more info about the way you are storing news.

final_task/rss_reader/classes/news_class.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+    def print_news(self):
+        """Print to stdout title, date, link, text
+        and links from text in particular sequence."""
+        print('-----------------------------------------------------------')
+        print('Title: '+self.title)
+        print('Date: '+self.date)
+        print('Link: '+self.link)
+        print()
```
HenadziStantchik: It is better to think of some other solution for printing an empty line instead of using `print()`

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,76 @@
...
+try:
+    import argparse
+    import functions.check_func as ch_f
+    import functions.process_func as proc_f
+    import functions.print_func as print_f
+    import classes.exceptions as exc
+    from functions.caching import cache_news, get_cached_news
+except ModuleNotFoundError as E:
+    print(E)
+    exit()
```
HenadziStantchik, line 30: It is better to have only one exit point from the application. Please rethink this solution

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,76 @@
...
+except KeyboardInterrupt:
+    print(" KeyboardInterrupt")
+    exit()
```
HenadziStantchik: You do not actually need to handle `KeyboardInterrupt` exception

final_task/setup.py
```diff
@@ -0,0 +1,25 @@
...
+    name="rss_reader_Anton_Zimahorau",
+    version="2.2",
+    author="Anton Zimahorau",
+    author_email="antonzimahorau@gmail.com",
+    description="Command-line utility which receives RSS URL",
+    long_description=long_description,
+    long_description_content_type="text/markdown",
+    url="https://github.com/AntonZimahorau/FinalTaskRssParser.git",
+    packages=['rss_reader', 'rss_reader.functions', 'rss_reader.classes'],
+    install_requires=['feedparser', 'lxml', 'argparse', 'requests'],
```
HenadziStantchik: Please make this correspond to your `requirements.txt` file

HenadziStantchik:
I have not discovered any issues with Iteration 3 work. Great job on implementing news extraction from local cache by source.
Resolve issues in comments and proceed to iteration 4.

final_task/rss_reader/rss_reader.py
AntonZimahorau: Why exit() here is unacceptable?

final_task/rss_reader/functions/check_func.py
AntonZimahorau: Fixed

final_task/rss_reader/functions/check_func.py
AntonZimahorau: Fixed

final_task/README.md
AntonZimahorau: Fixed

final_task/rss_reader/classes/news_class.py
AntonZimahorau: Fixed

final_task/rss_reader/rss_reader.py
AntonZimahorau: Fixed

final_task/setup.py
AntonZimahorau: Fixed

final_task/rss_reader/functions/convertation.py
```diff
@@ -0,0 +1,109 @@
...
+        content = str(doc_html)
+
+        chapter.set_content(content)
+        book.add_item(chapter)
+        book.spine.append(chapter)
+
+    book.add_item(epub.EpubNcx())
+    book.add_item(epub.EpubNav())
+
+    dir = os.path.join(command_line_args.to_epub, 'News.epub')
```
dzhigailo: Do not overwrite functions from stdlib. Please, rename `dir` variable

final_task/rss_reader/functions/convertation.py
```diff
@@ -0,0 +1,109 @@
...
+    try:
+        with open(dir, 'w') as h:
+            h.write("r")
+        epub.write_epub(dir, book)
+        logger.info("Epub book is created")
+    except FileNotFoundError:
```
dzhigailo: Instead of handling exception, you could check if file exists with `os.path.exists(path)`

final_task/rss_reader/functions/check_func.py
```diff
@@ -0,0 +1,54 @@
...
+import classes.exceptions as exc
+
+def check_internet_connection(logger):
+    """Check if network connection is avaliable"""
+    try:
+        logger.info("Checking internet connection...")
+        requests.get('https://www.google.com/', timeout=1)
+        return True
+    except requests.ConnectionError:
+        raise exc.InternetConnectionError("No internet access")
```
dzhigailo, line 16: Why would you want this function to raise exception? Following it's name, it should return either True or False

final_task/rss_reader/rss_reader.py
HenadziStantchik: Because you always want to know where in the code your application can finish its work.
It really helps with debugging and maintenance of large applications. And the easiest way to know all application exit points is to have only one such point: at the end of the main file.

HenadziStantchik:
1. HTML converted file has encoding problem:

zimahorau

2. I am not sure if this problem of my epub reader, but I did not have any images in epub converted file

3. There this exception when user tries to parse news from local storage to html without specifying rss url:

zimahorau

4. When parsing from local storage, images should be in the feed. If there is no internet connection, then it is ok to have links to them instead.

final_task/rss_reader/classes/news_class.py
```diff
@@ -22,17 +22,34 @@ def print_feed_title(self):
         """Print feed title"""
         print(self.feed_title)
 
+    def create_list_of_links(self):
+        """Turn tuple ([href_links], [img_links]) into list of links"""
+        # Merge list of links
+        list_of_links = []
+        list_of_links.extend(self.links[0])
+        list_of_links.extend(self.links[1])
+        return list_of_links
```
HenadziStantchik: A question:

If you absolutely sure that `self.links` corresponds to specified structure, why not use something like `self.links[0] + self.links[1]` instead of this function.

If there is possibility that `self.links` will have more than 2 elements, then it is better to use cycle and append to form a list.

final_task/rss_reader/classes/news_class.py
```diff
@@ -22,17 +22,34 @@ def print_feed_title(self):
...
+    def create_string_of_links(self):
+        """Turn tuple ([href_links], [img_links]) of lists into formatted string"""
+        # Merge list of links
+        list_of_links = self.create_list_of_links()
+
+        string_repr_of_links = ''
+        for num, link in enumerate(list_of_links):
+            if link:
+                string_repr_of_links = string_repr_of_links + '[{}] '.format(num + 1) + link + '\n'
+        self.links = string_repr_of_links
```
HenadziStantchik: It is better to not change existing object, but to return new object. This change is too implicit, and someone unfamiliar with your code will wonder why at some point of app work `self.links` has turned to a string.

final_task/rss_reader/classes/news_class.py
```diff
@@ -22,17 +22,34 @@ def print_feed_title(self):
...
+
+    def create_string_of_links(self):
+        """Turn tuple ([href_links], [img_links]) of lists into formatted string"""
+        # Merge list of links
+        list_of_links = self.create_list_of_links()
+
+        string_repr_of_links = ''
+        for num, link in enumerate(list_of_links):
+            if link:
+                string_repr_of_links = string_repr_of_links + '[{}] '.format(num + 1) + link + '\n'
```
HenadziStantchik: It is better to use `formatted string` here

final_task/rss_reader/functions/convertation.py
```diff
@@ -0,0 +1,109 @@
...
+    logger.info("Creating html file...")
+    doc_html = dominate.document(title='RSS News')
+    with doc_html:
+        tag.h1("RSS News")
+
+    # Internet connection checking
+    try:
+        internet = check_internet_connection(logger)
+    except InternetConnectionError:
+        internet = False
```
HenadziStantchik: You have code duplication here and above. It is better to move code to a separate function if you are using it more than one time, especially when you can easily make it.

final_task/rss_reader/functions/convertation.py
```diff
@@ -0,0 +1,109 @@
...
+
+    for num, news in enumerate(news_collection):
+        logger.info("Add entry {}".format(num + 1))
+        doc_html = create_html_news_entry(news, doc_html, internet)
+
+    dir = os.path.join(command_line_args.to_html, 'News.html')
```
HenadziStantchik: Same as Dzmitry commented above, it is a built-in name

final_task/rss_reader/functions/convertation.py
```diff
@@ -0,0 +1,109 @@
...
+    try:
+        with open(dir, 'w') as h:
```
HenadziStantchik: please avoid one-letter variable names

final_task/rss_reader/functions/check_func.py
AntonZimahorau: check_internet_connection() function invokes in main() function and should stop work of program, if we retrieve rss from internet source and internet is not avaliable. Well, in convertation.py file, it invokes to decide should we insert images links or not, and in this case I had to handle exception InternetConnectionError. I didn't changed check_internet_connection() function, because it would make main() function more complicated.

final_task/rss_reader/classes/news_class.py
AntonZimahorau: Fully agreed

final_task/rss_reader/classes/news_class.py
AntonZimahorau: It was conceived, that only two elements in this structure. Class method create_string_repr_of_links confused me.

final_task/rss_reader/functions/convertation.py
AntonZimahorau: Fixed

final_task/rss_reader/functions/convertation.py
AntonZimahorau: Fixed

final_task/rss_reader/functions/convertation.py
AntonZimahorau: Fixed

final_task/rss_reader/classes/news_class.py
AntonZimahorau: Fixed

final_task/rss_reader/classes/news_class.py
AntonZimahorau: Fixed

final_task/rss_reader/classes/news_class.py
AntonZimahorau: Fixed

final_task/rss_reader/functions/convertation.py
AntonZimahorau: Fixed

final_task/rss_reader/functions/convertation.py
AntonZimahorau: Fixed

final_task/rss_reader/classes/news_class.py
```diff
@@ -38,8 +36,8 @@ def create_string_of_links(self):
         string_repr_of_links = ''
         for num, link in enumerate(list_of_links):
             if link:
-                string_repr_of_links = string_repr_of_links + '[{}] '.format(num + 1) + link + '\n'
-        self.links = string_repr_of_links
+                string_repr_of_links = string_repr_of_links + "[{}] {}\n".format(num+1, link)
```
HenadziStantchik: It is still beter to use `formatetd string` here: `f"[{num+1}] {link}"`

final_task/rss_reader/functions/convertation.py
```diff
@@ -6,16 +6,108 @@
...
 
 
+def set_internet_flag(logger):
+    """If Internet connection, set flag to True, else False"""
+    try:
+        internet = check_internet_connection(logger)
+    except InternetConnectionError:
+        internet = False
+    finally:
+        return internet
```
HenadziStantchik: You do not really need to put this under `finally` block here. Is case of exception, the interpreter will run code inside the `except` block and then will run the code after it.

HenadziStantchik:

When parsing news entries to console (without any convertation) file `News.html` is created anyway. Is this intended?

Aside from that and some issues with your code I did not discover any more issues

final_task/rss_reader/functions/check_func.py
```diff
@@ -0,0 +1,54 @@
...
+    """If vebose argument is True, set logging to info level"""
+
+    logging.basicConfig(format='[%(asctime)s][%(levelname)s] %(message)s',
+                        stream=sys.stdout,
+                        level=logging.INFO
+                        )
+    logger = logging.getLogger("rss_reader_logger")
+    if command_line_args.verbose:
+        logger.setLevel(level=logging.INFO)
+        return logger
```
HenadziStantchik: you can move this `return` and `return` in `else` block to the end of the function

final_task/rss_reader/functions/convertation.py
```diff
@@ -0,0 +1,203 @@
...
+        if internet:
+            for image_object in list_of_image_objects:
+                tag.img(src=image_object.file_name)
+        else:
+            for link in images_links:
+                if link:
+                    tag.a("Image link", href=link)
+        tag.p(news.text)
+        for num, link in enumerate(all_links):
+            tag.a("Link №{}".format(num + 1), href=link)
```
HenadziStantchik: A minor improvement: use `formatted string` here: `f"..."`

final_task/rss_reader/functions/convertation.py
```diff
@@ -0,0 +1,203 @@
...
+    return list_of_images
+
+
+def create_image_objects(news, image_number):
+    """Return list of epubImage objects"""
+    list_of_image_objects = []
+    list_of_images = download_images(news)
+    for image in list_of_images:
+        img_obj = epub.EpubImage()
+        img_obj.file_name = "{}.jpg".format(image_number)
```
HenadziStantchik: Same here: `f"{image_number}.jpg"`

final_task/rss_reader/functions/print_func.py
```diff
@@ -0,0 +1,100 @@
+"""Contain formatting output functions"""
+
+import json
+import classes.exceptions as exc
+
+
+def check_limit_argument(command_line_args, news_collection, logger):
```
HenadziStantchik: It is better to give this function another name, which better describes what it does.

final_task/rss_reader/functions/print_func.py
```diff
@@ -0,0 +1,100 @@
...
+                    'Feed': news_collection[0].feed_title,
+                    'news': []
+                    }
+    logger.info("Configure json file...")
+    for news in news_collection:
+        news_dict = {}
+        news_dict['title'] = news.title
+        news_dict['date'] = news.date
+        news_dict['link'] = news.link
+        news_dict['text'] = news.text
```
HenadziStantchik: It is better to use method `asdict` from `dataclass` library for this. Check it out

final_task/rss_reader/functions/print_func.py
```diff
@@ -0,0 +1,100 @@
...
+    """Create json of news"""
+    all_news_dict = {'news': []}
+    logger.info("Configure json file...")
+    for news in news_collection:
+        news_dict = {}
+        news_dict['Feed'] = news.feed_title,
+        news_dict['title'] = news.title
+        news_dict['date'] = news.date
+        news_dict['link'] = news.link
+        news_dict['text'] = news.text
```
HenadziStantchik: Same as above: `asdict`

final_task/rss_reader/functions/convertation.py
AntonZimahorau: I know, I didn't want to write return internet two times

HenadziStantchik: And why would you need to write `return internet` two times?

```
try:
    internet = ...
except SomeException:
    internet = ...
return internet
```

final_task/rss_reader/functions/check_func.py
AntonZimahorau: Fixed

final_task/rss_reader/classes/news_class.py
AntonZimahorau: Fixed

final_task/rss_reader/functions/convertation.py
AntonZimahorau: Yes, i was wrong

final_task/rss_reader/functions/convertation.py
AntonZimahorau: Fixed

final_task/rss_reader/functions/convertation.py
AntonZimahorau: Fixed

final_task/rss_reader/functions/print_func.py
AntonZimahorau: Fixed

final_task/rss_reader/functions/print_func.py
AntonZimahorau: Fixed

AntonZimahorau: Well, when I was implementing fifth iteration, i was enforced to redefine asdict(), because News class became contain argparse.Namespace class instance. And I didn't need it in the dictionary.

final_task/rss_reader/functions/print_func.py
AntonZimahorau: Fixed

final_task/rss_reader/functions/print_func.py
HenadziStantchik: It is better to not redefine anything from standart libs, you could just `.pop()` anything you do not need from the dict

final_task/README.md
```diff
@@ -1,3 +1,92 @@
...
+
+### News storage
+>For news storage a “shelve”, persistent, dictionary-like object was used.
+>"Shelve object will be situated in your home directory. 
+>All fetched news will be saved in "shelve" object.
+>The keys for search is a rss publication date string. 
+>The values is python News() objects.
+>More information about "shelve" is avaliable on [link](https://docs.python.org/3/library/shelve.html).
```
HenadziStantchik, line 93: Please add info about 4th and 5th Iterations

final_task/rss_reader/functions/convertation.py
```diff
@@ -0,0 +1,204 @@
...
+        content = str(doc_html)
+
+        chapter.set_content(content)
+        book.add_item(chapter)
+        book.spine.append(chapter)
+
+    book.add_item(epub.EpubNcx())
+    book.add_item(epub.EpubNav())
+
+    path_to_ebook = os.path.join(command_line_args.to_epub, 'News.epub')
```
HenadziStantchik: It is better to have a bit more informative filename than this.

The same for html format

final_task/rss_reader/classes/news_class.py
```diff
@@ -40,16 +46,37 @@ def create_string_of_links(self):
...
+            print('-----------------------------------------------------------')
+            print('Title: '+self.title)
+            print('Date: '+self.date)
+            print('Link: '+self.link, end='\n\n')
+            print(self.text, end='\n\n')
+            links = self.create_string_of_links()
+            if links:
+                print('Links:')
+                print(links)
+            print('-----------------------------------------------------------')
```
HenadziStantchik: You have almost complete code duplication in these two statements. Try to rethink this logic. For example you could use a decorator here

final_task/rss_reader/functions/print_func.py
```diff
@@ -27,7 +28,7 @@ def generate_news_json(news_collection, logger):
     all_news_dict = {'news': []}
     logger.info("Configure json file...")
     for news in news_collection:
-        news_dict = asdict(news)
+        news_dict = news.asdict()
```
HenadziStantchik: better rename it to `as_dict` so it does not match with `asdict` from `dataclass` lib

final_task/rss_reader/classes/news_class.py
```diff
@@ -16,18 +18,22 @@ class News:
     links: str
     feed_title: str
     source: str
+    command_line_args: "argparse instance"
```
HenadziStantchik: It is better to try to think of some other method, instead of storing arguments in each instance of News class.

HenadziStantchik:
I did not discover any issues with Iteration 5 work

final_task/README.md
AntonZimahorau: Fixed

final_task/rss_reader/functions/convertation.py
AntonZimahorau: Fixed

final_task/rss_reader/classes/news_class.py
AntonZimahorau: Fixed

final_task/rss_reader/functions/print_func.py
AntonZimahorau: Fixed. Don't need it anymore

final_task/rss_reader/classes/news_class.py
AntonZimahorau: Fixed

final_task/rss_reader/functions/check_func.py
```diff
@@ -0,0 +1,54 @@
...
+def check_feed_status(feed):
+    """Check HTTP status code"""
+    try:
+        status = feed.status
+        if 500 > status >= 400:
+            raise exc.GettingFeedError("Client error – the request contains bad syntax or cannot be fulfilled.")
+        elif 600 > status >= 500:
+            raise exc.GettingFeedError("Server error – the server failed to fulfill an apparently valid request.")
+        else:
+            return feed
```
HenadziStantchik, line 44: It is better to use `bozo` attribute for this check.

`feed` here has type `FeedParserDict`, which is inherited from `dict`, so it is a dictionary, which has keys and values.
Neither `FeedParserDict` nor `dict` have attribute `status`, so theoretically you should always get an `AttributeError` here.
BUT `FeedParserDict` class has its magic method `__getattribute__` redefined (note: getattribute is called every time you try to access some object attribute using `.`)  in a way so if there is no such attribute (for example `status`), it will try to look in this dictionary keys, using `__getitem__` (note: getitem is called when you try to acccess some item in collection by key or by index: `data["status"]`). So there is no attribute `status`, but `status` is a key in `data` dictionary, so it is still luckily works.

In short, using `status` here is pretty volatile and may break your program.

HenadziStantchik:
Here is feedback on your app:
Iteration 1:
Everything works fine.
Iteration 2:
Everything installs fine. All iterations are working.
Iteration 3:
Everything works fine. Good job implementing search by rss source.
Iteration 4:
Everything works fine.
Iteration 5:
Everything works fine.

Overall code is readable, well structured, mostly corresponds to the PEP8 guidelines.
Tests are well written, coverage percentage is high enough.
Commit messages are mostly informative enough, but rarely correspond to the recommended commit message style.
Documentation could be bigger, it would be nice if you added some info about second iteration, for example how to install your package and etc.
Iteration 3 has improved utility with search by rss source, not just by date.

final_task/rss_reader/classes/news_class.py
```diff
@@ -0,0 +1,66 @@
...
+        if self.command_line_args.colorize:
+            cprint(self.feed_title, 'magenta')
+        else:
+            print(self.feed_title)
+
+
+    def print_news(self):
+        """Print to stdout title, date, link, text
+        and links from text in particular sequence."""
+        if self.command_line_args.colorize:
```
dzhigailo, line 52: Since this parameter will not change in runtime, there is no need to create `list_of_colors` and check `self.command_line_args.colorize`  every single time the function called. You could just check it upon object initialization and create proper `list_of_colors` and use it

PR #13, opened 2019-11-08
Vitali_Kozlou_h4j0rx@gmail.com

kaz1ou:
Iterations 1 and 2 are done. Please review.


final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,150 @@
...
+parser = argparse.ArgumentParser()
+parser.add_argument('--url', type=str, help='Please enter a valid URL for RSS Feed')
+parser.add_argument('--limit', type=int, default=1, help='You can set a limit for news')
+parser.add_argument('--verbose', type=bool, default=False, help='If you want to know '
+                                                                'what\'s happening, set this to True')
+parser.add_argument('--version', type=str, default=False, help='Print version of program')
+args = parser.parse_args()
+url = args.url
+verbose = args.verbose
+version = '[Reader version: 0.01]'
```
HenadziStantchik: It is better to put it into a function

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,150 @@
+import feedparser
+import argparse
+import html2text
+from requests import get
+from datetime import datetime
+
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--url', type=str, help='Please enter a valid URL for RSS Feed')
```
HenadziStantchik: RSS url is supposed to be passed to your application without option `--url` check example in FinalTask.md

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,150 @@
...
+
+def get_feed(url: str, verbose: bool):
+    """
+    Caching feed by creating a feedparser instance.
+    Returns non-formatted feed
+    :param url: url provided by user
+    :param verbose: True of False
+    :return: non-formatted feed
+    """
+    if verbose:
```
HenadziStantchik, line 68: Here and below. For such logging purposes it is bettert to use `logging` standart library. Check it out

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,150 @@
...
+    """
+    This function recieves feed and format all entries
+    by using them as to_text() arguments to remove
+    all HTML markup
+    :param feed: Cached feed returned by get_feed()
+    :param verbose: True or False
+    :param limit: Standart limit is 1 article
+    :return: title, entry_title(Article title), description(Text of article), date
+    """
+    global title, entry_title, description, date
```
HenadziStantchik: Here and below: It is better to avoid using global variables whenever you can. You can pass them as an arguments instead.

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,150 @@
...
+def main(title, entry_title, description, date):
+    """
+    This function just prints all parameters given by format_feed()
+    :param title: Title of whole feed returned by format_feed()
+    :param entry_title: Title of one article returned by format_feed()
+    :param description: Description of one article returned by format_feed()
+    :param date: Publishing date of one article returned by format_feed()
+    :return: None
+    """
+    print('[URL]: ' + args.url)
```
HenadziStantchik: It is better to make URL for each piece of news actual URL to the webpage containing this piece of news.

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,150 @@
...
+if __name__ == '__main__':
+    if args.url is None:
+        print('Please enter a valid URL for a feed '
+              'and run program like "python3 main.py '
+              '--url %YOUR FEED URL%')
+    if args.limit == 0:
+        print('You have entered zero values to print, please specify another limit value')
+    else:
+        feed = get_feed(url, verbose)
+        limit = [i for i in range(0, args.limit+1)]
```
HenadziStantchik: It is better to avoid one-letter variable names

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,150 @@
...
+    print('[Date]: ' + date)
+    return None
+
+
+if __name__ == '__main__':
+    if args.url is None:
+        print('Please enter a valid URL for a feed '
+              'and run program like "python3 main.py '
+              '--url %YOUR FEED URL%')
+    if args.limit == 0:
```
HenadziStantchik, line 147: It is better to move this line to the argument parsing

HenadziStantchik:
Some feedback on your application work:

1. If user did not specify `limit` then he will only get 1 news entry. He/she should get all available feed instead

2. If the limit is larger than number of feed entries then you get the `IndexError` with traceback in stdout.

3. There are no `--json` argument (check iteration 1)

4. `--verbose` option should not get any arguments. For example:
```python rss_reader.py <rss_url> --verbose```]
UPD: the same for `--version`

5. Please add logging to more functions.

6 (optional). A minor thing but if there are many news printed it becomes hard to tell where ends one and starts another. Please separate them a bit better.

**Comment by HenadziStantchik** [comment]
> Fixed all, according to your comments. Please review

Please do not use such commit messages. They should correspond to commit message style guide.
You can find it in `materials` section of our GIT session presentation

File: final_task/rss_reader/rss_reader.py [hunk 8]
```diff
@@ -0,0 +1,168 @@
...
+    :param url: url provided by user
+    :param verbose: True or False
+    :return: True in case of xml,
+    False and closing program in all other cases
+    """
+    if verbose:
+        print('Wait a second, we will check if there is any RSS in URL you have given')
+        print('-' * 50)
+    response = get(url)
+    r = response.headers['Content-Type']  # We will check only headers
```
HenadziStantchik, line 51: It is better to avoid one-letter variable names

HenadziStantchik:
1. If `--version` argument is specified application should just print its current bersion and shutdown.
Also user should be able to use `--version` argument without specifying source url.

2. Add a bit more logging to make ``-verbose` argument more explicit. For example, you can add some logging to functions that print news.

3. Right now your application just throws off images from the feed. for the future iterations you may want to have links to them (checkout example in FinalTask.md)

4. There is much redundant information in resulting json after using `--json` option.
There is also html and encoding problems in it.

PR #14, opened 2019-11-08
Victoria_Kondrat'eva_sam.kondrateva@gmail.com

Victoria-Sam:
Five iterations.


final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,143 @@
+import feedparser
+import sys
+import datetime
+import argparse
+import json
+from bs4 import BeautifulSoup
+
+DEFAULT_LIMIT_COUNT = 3
```
HenadziStantchik: If user did not specify `--limit` he/she should get all available feed. fith this line I suppose user will only get 3 entries.

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,143 @@
...
+VERSION = '1.0'
+
+
+class RSSReader:
+    is_verbose = False  # logging is disabled by default
+    is_json = False
+    limit = DEFAULT_LIMIT_COUNT
```
HenadziStantchik: If you are going to treat the whole reader as a class, it is better to make these instance attributes, instead of class attributes.

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def __init__(self, url):
+        self.url = url
+
+    def url_parsing(self):
+        parsed_url = feedparser.parse(self.url)
+        if parsed_url.status != 200:  # URL-access check
+            print("Can't load RSS feed.")
+            raise Exception('RSS-reader failed. Status: {}.'.format(parsed_url.status))
```
HenadziStantchik: Minor thing, but it is better to make your own custom exception here which inherits `Exception` class

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,143 @@
...
+                self.log("Error processing news: {} {}".format(type(ex), ex))
+                continue  # continues to process the following news
+            else:
+                all_news.append(dictionary_of_news_data)  # make list of dictionaries of news data for optional output
+        return all_news
+
+    def log(self, message):
+        """Print time and log message if verbose is active."""
+        if self.is_verbose:
+            print(datetime.datetime.now(), message)
```
HenadziStantchik: A creative solution, but nevertheless you should better check `logging` standart library. Your current method is hardly scalable and will cause problems for you in future iterations

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,143 @@
...
+        return json.dumps(dictionary, indent=4)
+
+    def output(self, about_website, all_news):
+        """Function which print information about site and a set of news."""
+        print()
+        for key, value in about_website.items():
+            print(key, ': ', value)
+            print()
+        for number_of_news in all_news:
+            print()
```
HenadziStantchik: If you need empty prints just for empty line in the output it is better to try to do it some other way.

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,143 @@
...
+        except Exception as ex:  # for any exception after website parsing
+            self.log("Error reading site data: {}, {}".format(type(ex), ex))
+            return
+        string_of_news_dictionaries = self.news_data_collection(parsed_url)
+        if self.is_json:
+            self.log('Convert to JSON-format')
+            self.output_json(about_website, string_of_news_dictionaries)
+        else:
+            self.log('Output news')
+            self.output(about_website, string_of_news_dictionaries)
```
HenadziStantchik: It is better to have a separate module for the each class you have.

HenadziStantchik:

1. If user does not specify `limit` he/she should get all available feed.

2. Your app has encoding problems. (check the image below)

3. Sometimes there is no link in the `Source of image`

image

4. For future iterations it wold be better to print where exactly the image is inside the news text (check the example in `FinalTask.md`)

5 (optional). It would be nice if you add some more logging.

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def make_news_data(self, news):
+        """
+        Function which parsed HTML summary of news and make dictionary with all the necessary news data.
+        :return dictionary
+        """
+        bs = BeautifulSoup(news['summary'], 'html.parser')
+        img = bs.find_all('img')
+        image_data = 'No image'
+        if len(img) > 0:
```
dzhigailo: Such check should look: `if not img:`

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,143 @@
...
+        all_information_about_news = parsed_url['entries']
+        # amount_of_news = len(parsed_url.entries)
+        all_news = []
+        self.log('News gathering')
+        for news in all_information_about_news[:self.limit]:
+            try:  # Exception when news is failed
+                dictionary_of_news_data = self.make_news_data(news)
+            except Exception as ex:
+                self.log("Error processing news: {} {}".format(type(ex), ex))
+                continue  # continues to process the following news
```
dzhigailo: You dont need `continue` in here. After handling of an excetion, cycle will continue anyway

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,143 @@
...
+            print()
+        for number_of_news in all_news:
+            print()
+            for key, value in number_of_news.items():
+                print(key, ': ', value)
+
+    def output_json(self, about_website, all_news):
+        print(self.parse_to_json([about_website] + all_news))
+
+    def doing_everything_in_class(self):
```
HenadziStantchik: It is better to use different method name here

final_task/README.md
```diff
@@ -1,3 +1,35 @@
...
+        "Date": "Thu, 07 Nov 2019 09:59:52 -0500",
+        "Summary": "Some summary",
+        "Link": "Link to news",
+        "Image": "Some image information"
+    }
+]
+```
+
+The following external libraries are used in this program:
+feedparser and bs4.
```
HenadziStantchik: Please add info about how do you store cached news and how to launch your program using setuptools

final_task/rss_reader/NewsCache.py
```diff
@@ -0,0 +1,42 @@
...
+        self.logger = logging.getLogger(__name__)
+        self.file_name = file_name
+        self.logger.info("Check for cache file...")
+        if os.path.isfile(file_name):
+            try:
+                with open(file_name, "r") as json_file:
+                    self.cache_news = json.load(json_file)
+                    self.logger.info("Reading cache file...")
+            except Exception as ex:
+                self.logger.error("Error reading file: {} {}".format(type(ex), ex))
```
HenadziStantchik, line 21: It is better to use `os.path.exists` instead of this

final_task/rss_reader/NewsCache.py
```diff
@@ -0,0 +1,42 @@
...
+                self.cache_news = {}
+        else:
+            self.logger.warning("Warning! Caching file is not found!")
+            self.cache_news = {}
+
+    def caching(self, all_news, url):
```
HenadziStantchik, line 27: Here and below: please use more informative method names

final_task/rss_reader/RSSReader.py
```diff
@@ -0,0 +1,130 @@
...
+        """
+        Function which parsed HTML summary of news and make dictionary with all the necessary news data.
+        :return dictionary
+        """
+        if 'summary' in news:
+            bs = BeautifulSoup(news['summary'], 'html.parser')
+            img = bs.find_all('img')
+            image_data = 'No image'
+            if img:
+                image_data = '{}\nSource of image: {}'.format(img[0].get('alt', ""), img[0]['src'])
```
HenadziStantchik: Here and below: It is better to use `formatted string` instead of `.format`

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,44 @@
...
+    level = logging.INFO if args.verbose else logging.ERROR
+    logging.basicConfig(format='[%(asctime)s][%(levelname)s]%(message)s', stream=sys.stdout, level=level)
+    if args.version:
+        print('RSS-reader version {}'.format(VERSION))  # program version call
+        return
+    reader = RSSReader()
+    reader.date = args.date
+    reader.url = args.source
+    reader.limit = args.limit
+    reader.is_json = args.json
```
HenadziStantchik: You can move this logic inside RSSReader `__init__` method

HenadziStantchik:
1. User should be able to specify RSS url without `--source` option

2. Your app has encoding problem

kondrat

final_task/rss_reader/NewsCache.py
Victoria-Sam: But above there is os.path.isfile. Why os.path.exists is better?

HenadziStantchik: You won't need to handle `FileNotFoundError` in this case

HenadziStantchik:
Pdf conversion seems bugged:

k

Keep in mind that user should be able to specify the path where converted file should be saved.

final_task/rss_reader/NewsCache.py
Victoria-Sam: This exception is for such error like nonreadable format of file and in this case file will be overwritten.

final_task/setup.py
```diff
@@ -0,0 +1,18 @@
+import sys
+sys.path.append('rss_reader')
+from setuptools import setup, find_packages
+from rss_reader import rss_reader
```
HenadziStantchik, line 4: This will not work on clean machine, if there are no libs from `requirements.txt` installed.

final_task/setup.py
```diff
@@ -0,0 +1,18 @@
...
+
+setup(
+    name='rss_reader',
+    version=rss_reader.VERSION,
```
HenadziStantchik, line 8: In this specific case it is better either hardcode this value or import in from some file.

final_task/setup.py
```diff
@@ -0,0 +1,18 @@
...
+    packages=find_packages(),
+    install_requires=['feedparser', 'bs4', 'unidecode', 'fpdf', 'ebooklib', 'colored'],
+    author="Victoria Kondrat'eva",
+    author_email='sam.kondrateva@gmail.com',
+    url='https://github.com/Victoria-Sam',
+    description='RSS Reader',
+    entry_points={'console_scripts': ['Rss-Reader = rss_reader.rss_reader:main']},
+    keywords="rss-reader",
+    python_requires='>=3.7',
```
HenadziStantchik, line 17: It should be 3.8

HenadziStantchik:
Iteration 1:
Everything works fine
Iteration 2:
Does not install.
image
Iteration 3:
Everything works fine.
Iteration 4:
EPUB file does not contain images.
Keep in mind that user must specify a directory, not a file in `--to_pdf` and `--to_epub` options
Iteration 5:
Everything works fine.

Tests are ok, but the coverage is barely enough.
Project documentation is decent.
The code overall is readable, absolutely corresponds to the PEP8 guidelines and decently structured.
Commit messages could be more informative, but correspond to the recommended commit message style.

final_task/setup.py
Victoria-Sam: I tried running on a clean machine.  When libraries is installed, just start it as rss_reader.rss_reader.
Example:
python -m rss_reader.rss_reader "http://news.yahoo.com/rss/" --limit 2 --verbose

final_task/rss_reader/output.py
```diff
@@ -0,0 +1,143 @@
...
+import json
+import urllib
+from colored import attr, fg, bg
+from ebooklib import epub
+from fpdf import FPDF
+
+
+def get_output_function(converter):
+    """Output factory"""
+    if converter == 'json':
```
dzhigailo, line 11: Instead if multiple if statements you could create dictionary and get needed handler by key, e.g
{
  'json': json_handler,
....
}

final_task/setup.py
HenadziStantchik: > When libraries is installed


Yes, _when_ libraries are installed. The idea of second iteration is that your app will download required libs itself, without you doing `pip install -r requirements.txt`

> python -m rss_reader.rss_reader "http://news.yahoo.com/rss/" --limit 2 --verbose

Your app works perfectly while using `python rss_reader...`, But the idea of second iteration is to install a CLI utility, that will allow you to run your app like this:
```
rss-reader <rss_source> <options>
```
Without manually installing required libs.

PR #15, opened 2019-11-09
Oleg_Slavashevich_oslavashevish@gmail.com

OlegSlavashevich:


final_task/rss_reader/myargparse.py
```diff
@@ -0,0 +1,20 @@
+import argparse
+
+parser = argparse.ArgumentParser(description='Getting info from sites')
```
dzhigailo: Entire logic of this module should placed in a function.

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,61 @@
+import feedparser
+from pprint import pprint
+from myargparse import *
```
dzhigailo: Please, do not use `import *` in your solutions

final_task/rss_reader/myargparse.py
OlegSlavashevich: Ok

final_task/rss_reader/rss_reader.py
OlegSlavashevich: Ok

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,61 @@
...
+from bs4 import BeautifulSoup 
+import json
+
+
+def parse(url):
```
dzhigailo: I'm not sure that you have to add this function. All it does - return result of another function.

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,60 @@
+import feedparser
+from pprint import pprint
+from myargparse import parsargs
```
HenadziStantchik: It is preferrable to avoid names such as `my....` in the production code

HenadziStantchik:
Please check your code for PEP8 compliance. You can use such tools as `pycodestyle` (but make maximum line length to 120 when you check)

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,60 @@
...
+    print("\nImage: ", value['img'])
+    print('\n')
+
+if __name__ == '__main__':
+    feed = get_sourse(parsed)
+    articles = get_news(parsed)
+    print('Feed: ', feed['link'], '\n')
+    if args.json:
+        for value in articles:
+            j = to_json(articles)
```
HenadziStantchik: It is better to avoid one-letter variable names whenever possible.

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,60 @@
...
+    """ Convert to json """ 
+    return json.dumps(articles)
+
+def output(articles):
+    print("Title: ", value['title'])
+    print("Date: ", value['published'])
+    print("Link: ", value['link'])
+    print("\nSummary: ", value['summary'])
+    print("\nImage: ", value['img'])
+    print('\n')
```
HenadziStantchik: It is better to pass value as an argument here, not the whole articles list.

HenadziStantchik:

1. There should not be any HTML-code in the printed feed.

2. Your application has encoding problem:

image

3. If user did not specify the limit, he/she should get all available feed, not just one entry.

Please do not forget that you have to add `--version` and  `--verbose` arguments in Iteration 1.

final_task/rss_reader/loggs.log
```diff
@@ -0,0 +1,22 @@
+INFO:root:Website is working
```
dzhigailo, line 1: Please, remove your local .log file from repository

final_task/rss_reader/loggs.py
```diff
@@ -0,0 +1,15 @@
+import logging
+from clean_output import clean_title
```
dzhigailo: This is an unused import

final_task/rss_reader/loggs.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+logging.basicConfig(filename="loggs.log", level=logging.DEBUG)
+
+
+def logg(article):
+    logging.debug("Title: " + clear_title(article['title']))
```
dzhigailo: This code is not working, because there is no clear_title function

final_task/rss_reader/arg.py
```diff
@@ -0,0 +1,36 @@
...
+        '--verbose',
+        action="store_true",
+        help='Outputs verbose status messages'
+    )
+
+    args = parser.parse_args()
+    
+    return args
+
+vers = 1.0
```
HenadziStantchik: Global constants are usually declared in upperacase, e.g. `VERSION`

final_task/rss_reader/clean_output.py
```diff
@@ -0,0 +1,8 @@
+from string import ascii_letters, whitespace
+import re
+
+def clean_title(text):
+    "Delete unnecessary symbols"
```
HenadziStantchik, line 5: Here and below: please make your docstrings a bit more specific

final_task/rss_reader/clean_output.py
```diff
@@ -0,0 +1,8 @@
+from string import ascii_letters, whitespace
+import re
+
+def clean_title(text):
```
HenadziStantchik: Function name is a bit misleading.

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,76 @@
...
+from arg import parsargs, vers
+from clean_output import clean_title
+from bs4 import BeautifulSoup 
+import json
+from loggs import logg, logg_json
+import logging
+
+args = parsargs()
+
+parsed = feedparser.parse(args.source)
```
HenadziStantchik: It is better to avoid using global variables whenever you can. For example here you can pass args as an argument to a function if you need,

final_task/setup.py
```diff
@@ -0,0 +1,11 @@
...
+from os.path import join, dirname
+
+setup(
+    name='RSS-reader',
+    version='1.0',
+    packages=find_packages(),
+    long_description=open(join(dirname(__file__), 'README.md')).read(),
+    author="Oleg Slavashevich",
+    author_email="oslavashevish@gmail.com"
+)
```
HenadziStantchik: You did not specify required packages for your app. Also you don't have `requirements.txt` file. Did you check that your second iteration at least works?

HenadziStantchik:
1. `--verbose` does not work: I get a `NameError` exception with traceback to stdout.

2. it is impossible to use `--version` without specifying RSS url

HenadziStantchik:
Please fix your commit messages as some of them do not correspond to commit message guidelines.

HenadziStantchik:
Please thoroughly test your app before pushing your commits to the remote repository.

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,76 @@
+import feedparser
+from pprint import pprint
```
HenadziStantchik: unused import here

final_task/rss_reader/arg.py
```diff
@@ -0,0 +1,42 @@
...
+def parsargs():
+    parser = argparse.ArgumentParser(description='Getting info from sites')
+
+    parser.add_argument('source', type=str, help='RSS URL')
+
+    parser.add_argument(
+        '--limit',
+        type=int, 
+        help='Limit news topics if this parameter provided'
+    )
```
HenadziStantchik, line 12: Here and below: it is ok to make that into one line. It will be readable anyway.

final_task/rss_reader/converter.py
```diff
@@ -0,0 +1,19 @@
...
+             'Jun': '6',
+             'Jul': '7',
+             'Aug': '8',
+             'Sep': '9',
+             'Oct': '10',
+             'Nov': '11',
+             'Dec': '12'}
+    day = date[5:7]
+    month_int = month[date[8:11]]
+    year = date[12:16]
```
HenadziStantchik, line 17: It is better to use `datetime` lib for this.

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,107 @@
+import feedparser
+#from arg import parsargs, VERSION
```
HenadziStantchik, line 2: please avoid commented-out code in pushed commits.

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,107 @@
...
+    print("Link: ", article['link'])
+    print("\nSummary: ", article['summary'])
+    print("\nImage: ", article['img'])
+    print('\n')
+
+def test_to_add(news_csv, articles):
+    for i in articles:
+        var = True
+        for j in news_csv:
+            if i['published'] != j[4]:
```
HenadziStantchik, line 56: It is better to avoid one-character variable names.

final_task/rss_reader/add_to_csv.py
```diff
@@ -0,0 +1,17 @@
+import csv
+
+def addcsv(articles):
+    with open('news.csv', 'a', newline='') as csvfile:
+        fieldnames = ['link','title', 'img', 'summary', 'published']
+
+        writer = csv.DictWriter(csvfile, fieldnames=fieldnames)
+
+        writer.writerow(articles)
```
HenadziStantchik, line 9: As far as I understand if you will get news 5 times from the same rss source, then you will have these news 5 times repeated in your file

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,107 @@
...
+                var = True
+            else:
+                var = False
+                break
+        if var:
+            addcsv(i)
+
+def main():
+    console_args = arg.parsargs()
```
HenadziStantchik, line 65: If you import specific function using `from` then you should just use it as is: `parsargs`, without `arg`

Right now this will break your application.

final_task/rss_reader/add_to_csv.py
```diff
@@ -0,0 +1,17 @@
...
+
+def out():
+    a = []
+    with open('news.csv', 'r') as csvFile:
+        reader = csv.reader(csvFile)
+        for row in reader:
+            a.append(row)
```
HenadziStantchik, line 16: If there is no such file (and the first time you launch program on new machine there won't be such file) this line will crash your application with `FileNotFoundError`

HenadziStantchik:
Here is feedback on your app:
Unfortunatelly your app does not work:
image


There are no tests.
The code is mostly readable, but does not correspond to the PEP8 guidelines. You may want to use `pycodestyle` tool in the future.
Commit messages could be more informative, however they mostly correspond to the recommended commit message style.

PR #16, opened 2019-11-09
Anna_Gonchar_raphaelkyzy@gmail.com

AnnaPotter:


final_task/rss_reader/action_functions.py
```diff
@@ -0,0 +1,118 @@
...
+    news_collection = {}
+    feed = {"title": news_feed.feed.get("title", ""),
+            "date": news_feed.feed.get("published", ""),
+            "language": news_feed.feed.get("language", "")}
+
+    news_collection["feed"] = feed
+    news_collection["entries"] = []
+
+    for entry in news_feed.entries:
+        news_entry = NewsEntry()
```
dzhigailo, line 104: There is a misunderstanding of what attributes do you assign. Your class NewEntry has class attribute `title`, `date` and so on. You are working not with them, after creating of the instance of class NewEntry. You create new attributes `title`, `date`, `link`, `summary`, which reffer to instance. You never change class attributes. In order to make things work properly, please take a look at `dataclasses`.

final_task/rss_reader/models.py
```diff
@@ -0,0 +1,48 @@
+""" Data models module """
+import argparse
+
+
+class ArgReader:
```
dzhigailo: Why do you need a wrapping class for `parser`? You could just make a function just like __init__() method here and return parser object

final_task/rss_reader/models.py
```diff
@@ -0,0 +1,48 @@
...
+
+        Methods:
+        print_entry(self) - print entry in stdout """
+
+    title: str
+    summary: str
+    date: str
+    link: str
+
+    def print_entry(self):
```
dzhigailo, line 22: Above and below:
Why do you use prints, when you created a good logger object?

final_task/rss_reader/action_functions.py
```diff
@@ -0,0 +1,118 @@
...
+        news_collection["entries"].append(news_entry)
+
+    return news_collection
+
+
+def print_news_stdout(news_collection):
+    """ Function for print news to stdout in text format. """
+    print("###############################################################")
+    print()
+    print("Feed: ", news_collection["feed"]["title"])
```
HenadziStantchik: If you need an empty line, there are much better solutions to this. It is better not to use `print()` without arguments

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,19 @@
...
+logger = action_functions.create_logger(com_line_args)
+
+validation_functions.check_version_arg(com_line_args, logger)
+validation_functions.check_internet_connection(com_line_args, logger)
+validation_functions.check_url(com_line_args, logger)
+
+news_collection = action_functions.get_news(com_line_args, logger)
+validation_functions.check_emptiness(news_collection, logger)
+
+action_functions.print_news(news_collection, com_line_args, logger)
```
HenadziStantchik: It is better to put it inside `if __name__ == '__main__':`
It will help you a bit on second Iteration

final_task/rss_reader/validation_functions.py
```diff
@@ -0,0 +1,82 @@
...
+    """ Check URL function. """
+    req = Request(com_line_args.source)
+    try:
+        logger.info("Checking url.")
+        response = urlopen(req)
+    except URLError as e:
+        if hasattr(e, "reason"):
+            logger.error(f"Failed to reach a server. ")
+            print("Please, check your URL.")
+            exit()
```
HenadziStantchik: It is better to have only one exit point from the application. For example, you can raise your custom exception here and handle it in the main function

final_task/rss_reader/validation_functions.py
```diff
@@ -0,0 +1,82 @@
...
+    else:
+        return True
+
+
+def check_version_arg(com_line_args, logger):
+    """ Check --version argument function. """
+    if com_line_args.version:
+        logger.info("View program version.")
+        print("rss_reader.py 1.0")
+        exit()
```
HenadziStantchik: Same as above about application exit point

final_task/rss_reader/validation_functions.py
```diff
@@ -0,0 +1,82 @@
...
+        Analyzes the received value and makes it valid for correct program work. """
+    if com_line_args.limit or com_line_args.limit == 0:
+        limit = com_line_args.limit
+    else:
+        limit = len(news_collection["entries"])
+    if limit < 0:
+        logger.warning("Command line argument limit is invalid.")
+        print("Command line argument limit is invalid. "
+              "It should not be negative.")
+        com_line_args.limit = int(input("Please, enter a valid value:"))
```
HenadziStantchik: Our application is supposed to be 'one-shot' so if user did not specify correct `--limit`, just shutdown the program.

final_task/rss_reader/json_structure.txt
```diff
@@ -0,0 +1,20 @@
...
+                 "link": news link },
+
+               { "title": news title,
+                 "summary": news content,
+                 "date": news publication date,
+                 "link": news link },
+
+               ...
+             ]
+}
```
HenadziStantchik: It is better to move it into `final_task/README.md`. All technical info is supposed to be there.

HenadziStantchik:

1. There should not be any HTML code in the printed news (check the example and image below)
Same for the JSON format

2. Your app has encoding problem (check the image below)

ma

3. User should be able to get the app version without specifying RSS source. For example like this:
```python rss_reader.py --version```

File: final_task/rss_reader/models.py (hunk 3 above)
Comment (AnnaPotter) [reply final_task/rss_reader/models.py:22]: As far as I understand this question isn't related to part of code above, am I right?

File: final_task/rss_reader/action_functions.py (hunk 1 above)
Comment (AnnaPotter) [reply final_task/rss_reader/action_functions.py:104]: fixed

File: final_task/rss_reader/models.py (hunk 2 above)
Comment (AnnaPotter) [reply final_task/rss_reader/models.py]: fixed

File: final_task/rss_reader/action_functions.py (hunk 4 above)
Comment (AnnaPotter) [reply final_task/rss_reader/action_functions.py]: fixed

File: final_task/rss_reader/rss_reader.py (hunk 5 above)
Comment (AnnaPotter) [reply final_task/rss_reader/rss_reader.py]: fixed

File: final_task/rss_reader/validation_functions.py (hunk 6 above)
Comment (AnnaPotter) [reply final_task/rss_reader/validation_functions.py]: fixed

File: final_task/rss_reader/validation_functions.py (hunk 7 above)
Comment (AnnaPotter) [reply final_task/rss_reader/validation_functions.py]: fixed

File: final_task/rss_reader/validation_functions.py (hunk 8 above)
Comment (AnnaPotter) [reply final_task/rss_reader/validation_functions.py]: fixed

File: final_task/rss_reader/json_structure.txt (hunk 9 above)
Comment (AnnaPotter) [reply final_task/rss_reader/json_structure.txt]: fixed

File: final_task/rss_reader/action_functions.py [hunk 10]
```diff
@@ -72,12 +87,12 @@ def get_news(command_line_args, logger):
 
 def print_news_stdout(news_collection):
     """ Function for print news to stdout in text format. """
-    print("###############################################################")
-    print()
+    print("################################################################################")
+    print("")
```
HenadziStantchik: It is still better to look for a better solution than `print("")`

HenadziStantchik:
> 1. There should not be any HTML code in the printed news (check the example and image below)
>    Same for the JSON format
> 2. Your app has encoding problem (check the image below)
> 
> ma

HenadziStantchik:
Iteration 1 does not seem to have issues anymore. Proceed to the next Iteration

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,24 @@
...
+        validation_functions.check_internet_connection(com_line_args, logger)
+        validation_functions.check_url(com_line_args, logger)
+
+        news_collection = action_functions.get_news(com_line_args, logger)
+        validation_functions.check_emptiness(news_collection, logger)
+
+        action_functions.print_news(news_collection, com_line_args, logger)
+    except exceptions.Error as e:
+        print(e)
+        exit()
```
HenadziStantchik: You do not really need `exit()` here. After exiting from this exception block script will finish its work naturally.

final_task/rss_reader/validation_functions.py
```diff
@@ -0,0 +1,76 @@
...
+
+def check_internet_connection(com_line_args, logger):
+    """ Check internet connection function. """
+    try:
+        logger.info("Checking internet connection.")
+        response = requests.get("http://google.com", timeout=5)
+        return True
+    except requests.exceptions.ConnectionError:
+        logger.error("No internet connection.")
+        raise Error("Please, check your internet connection.")
```
HenadziStantchik: The function above does almost the same. Do you really need this one?

final_task/rss_reader/validation_functions.py
```diff
@@ -0,0 +1,76 @@
...
+    else:
+        return True
+
+
+def check_version_arg(com_line_args, logger):
+    """ Check --version argument function. """
+    if com_line_args.version:
+        logger.info("View program version.")
+        print("rss_reader.py 1.0")
+        raise Error("")
```
HenadziStantchik: It is better to not raise errors to deliberately finish the program. Please look for some other solution

final_task/rss_reader/rss_reader.py
AnnaPotter: fixed

final_task/rss_reader/validation_functions.py
AnnaPotter: fixed

final_task/rss_reader/validation_functions.py
AnnaPotter: I use two functions to tell what the problem is more precisely and to give more specific instructions for its correction. Do you think that it is unjustified to use an additional library for just one check?

final_task/rss_reader/exceptions.py
```diff
@@ -0,0 +1,10 @@
+""" Module for non-standard exceptions. """
+
+
+class Error(Exception):
+    """ Class to raising exceptions. """
+    def __init__(self, message):
+        super().__init__(message)
+
```
HenadziStantchik: You don't really need this logic here. If you leave `pass` here it will do the same thing

final_task/rss_reader/validation_functions.py
HenadziStantchik: But you will not be able to use urlopen in the first function without internet connection anyway. 
Why not bundle it together if it does the same thing and does it in a similar way?

final_task/setup.py
```diff
@@ -0,0 +1,28 @@
...
+    url="https://github.com/AnnaPotter/FinalTaskRssParser",
+    author="Anna Gonchar",
+    author_email="raphaelkyzy@gmail.com",
+    packages=find_packages(),
+    classifiers=[
+        "Programming Language :: Python :: 3.8",
+        "Operating System :: OS Independent",
+    ],
+    python_requires='>=3.8',
+    install_requires=['feedparser', 'requests', 'beautifulsoup4==4.8.1'],
```
HenadziStantchik: this line and your `requirements.txt` do not match

AnnaPotter: fixed

final_task/rss_reader/exceptions.py
AnnaPotter: fixed

final_task/rss_reader/validation_functions.py
AnnaPotter: I just don’t like the fact that the exception in urlopen both in the absence of the Internet and in the case of an incorrectly entered url is the same, albeit with different reasons. It is immediately clear that there is no Internet connection, in the case of request. But if necessary, I will fix it, leaving only the first function)

HenadziStantchik: It is not that critical

final_task/rss_reader/action_functions.py
AnnaPotter: fixed

AnnaPotter:
README.md will be fixed in the next commit.

final_task/rss_reader/action_functions.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+def print_news(news_collection, com_line_args, logger):
+    """ Function for print news to stdout,
+        that take account of limit and json arguments. """
+
+    # get valid limit argument
+    # if not initialize limit argument
+    if not check_limit_arg(com_line_args, logger):
+        limit = len(news_collection["entries"])
```
dzhigailo: In that case, you may want to set limit to None and refactor your logic below.

final_task/rss_reader/action_functions.py
```diff
@@ -0,0 +1,195 @@
...
+    else:
+        limit = com_line_args.limit
+
+    if len(news_collection["entries"]) < limit:
+        logger.warning("The number of news is less than the value of the argument limit.")
```
dzhigailo: To my mind, it's not the warning that user may want to see. For example, we have only 3 news and the limit is 5. It's completly ok. You could notify user, available news > chosen limit. But this is optional

AnnaPotter: This warning only for me, users can see only errors or higher level logs.

final_task/rss_reader/action_functions.py
AnnaPotter: fixed

HenadziStantchik:
1. It there are no images in the article, `Images links` are printed empty:
image

2. `-h` is bugged:
image

3. Keep in mind that converted file **should contain images**, not just links to them.

4. Pdf conversion does not work:
image

final_task/rss_reader/action_functions.py
```diff
@@ -0,0 +1,179 @@
...
+    create_logger(com_line_args) -> logger
+    get_com_line_args() -> com_line_args
+    get_news(command_line_args, logger) -> news_collection
+    print_news_stdout(news_collection) -> None
+    print_news_json(news_collection) -> None
+    print_news(news_collection, com_line_args, logger) -> None
+    print_cache_news(news_collection, logger) -> None
+    print_cache_news_json(news_collection, logger) -> None
+    convert_date(date_str, logger) -> str_date
+    clean_str(string) -> clean_string """
```
HenadziStantchik, line 13: Here and below: there is actually no need to write function list in module docstring.

final_task/rss_reader/conversion_functions.py
```diff
@@ -0,0 +1,148 @@
...
+    pdf.write(10, "News publication date: ")
+    pdf.set_text_color(0, 0, 0)
+    pdf.multi_cell(0, 10, txt=f"{news.date}")
+    pdf.set_text_color(0, 0, 255)
+    pdf.write(10, "Summary: ")
+    pdf.set_text_color(0, 0, 0)
+    pdf.multi_cell(0, 10, txt=f"{news.summary}")
+    pdf.set_text_color(0, 0, 255)
+    pdf.write(10, "News link: ")
+    pdf.set_text_color(0, 0, 0)
```
HenadziStantchik, line 70: It is better to think of a way to simplify this.

HenadziStantchik:
Iteration 1:
Everything works fine.
Iteration 2:
Installs, but does not work:
image
Iteration 3:
Does not work:
image
Iteration 4:
Everything works fine

Tests are decently written. Coverage percentage is high enough.
The code overall is readable, absolutely corresponds to the PEP8 guidelines.
Commit messages could be a bit more informative. Also they mostly correspond to the recommended commit message style.

AnnaPotter:
Thanks a lot for review.  A little annoying for the third iteration, because I checked. But now it really does not work... Thanks again)

final_task/rss_reader/caching_functions.py
```diff
@@ -0,0 +1,65 @@
...
+
+        if not check_limit_arg(com_line_args, logger):
+            limit = len(news_dict)
+        else:
+            limit = min(com_line_args.limit, len(news_dict))
+
+        if source:
+            for hash_date_key in news_dict:
+                if date in hash_date_key:
+                    if hash_date_key.split()[1] == date.partition(' ')[0]:
```
HenadziStantchik, line 45: This line (and the same line below) breaks your date search algorithm for days of month that start from 0, e.g. 01,02,03.
For example for specified `--date 20191203` in this function:
```
date = 3 Dec 2019
hash_date_key = 'Tue, 03 Dec 2019 07:11:16 -0500'
date in hash_date_key -> True

BUT:
hash_date_key.split()[1] -> '03'
and
date.partition(' ')[0] -> '3'
```
I actually noted that your Iteration 3 worked before final review.

HenadziStantchik:
> A little annoying for the third iteration, because I checked. But now it really does not work...

After a bit of investigation, check my comment about `caching_functions.py` above

AnnaPotter:
You are right, it was these two lines that broke the whole iteration. It’s just that the dates came in the format (Publication date: Wed, 5 Feb 2014 09:00:00 -0500) on my news feed and my previous version of the function with parameter --date 20190205 / 2019025, for example, did not work correctly. It displayed either the lack of news, or the news from 20190215/20190225. I tried to fix this bug, but I did not think through it thoroughly. Next time I will be more attentive.

PR #20, opened 2019-11-14
-

SergeyRogonov:
The first iteration with its all functionality
add parser.py, news_articles.py, .gitignore
changed rss_reader.py, requirements.txt


HenadziStantchik:
Please fix this PR's name so it corresponds the one in README.md

final_task/rss_reader/requirements.txt
```diff
@@ -0,0 +1,14 @@
...
+isort==4.3.21
+lazy-object-proxy==1.4.3
+lxml==4.4.1
+mccabe==0.6.1
+pycodestyle==2.5.0
+pyflakes==2.1.1
+pylint==2.4.3
+six==1.13.0
+soupsieve==1.9.5
+wrapt==1.11.2
```
HenadziStantchik: There are many libs that you do not use here. It is better to clean this up.

HenadziStantchik:

1. when using `--json` option, your app has encoding problem:

rogonov

2. If `--limit` is larger than feed size, `StopIteration` is printed with traceback to stdout.

final_task/rss_reader/news_articles.py
```diff
@@ -0,0 +1,29 @@
+
+
+class NewsArticle:
+    """Contain data about an article"""
+    def __init__(self, news_outlett_name, news_title, pub_date, news_link,
```
HenadziStantchik: Consider using `dataclass` here

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Pure python command-line RSS reader")
+    parser.add_argument("source", help="RSS URL")
+    parser.add_argument("--version", help="Print version info", action="version", version='rss_reader 1.0')
+    parser.add_argument("--json", help="Print result as JSON in stdout", action="store_true")
+    parser.add_argument("--verbose", help="Outputs verbose status messages", action="store_true")
+    parser.add_argument("--limit", help="Limit news topics if this parameter provided")
+    args = parser.parse_args()
```
HenadziStantchik: Better move this to a separate function

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,72 @@
...
+    """getting data"""
+    try:
+        response = request.urlopen(url)
+    except error.HTTPError as ex:
+        logging.info(ex)
+        raise
+    return response
+
+
+def output_format(news_articles, json_output=False):
```
HenadziStantchik: Please be consistent with typehints. It is better to either typehint everywhere(preferrable) or nowhere(still ok)

final_task/rss_reader/requirements.txt
SergeyRogonov: fixed

final_task/rss_reader/news_articles.py
SergeyRogonov: fixed. it really looks better

final_task/rss_reader/rss_reader.py
SergeyRogonov: fixed

final_task/rss_reader/rss_reader.py
SergeyRogonov: removed it for now

final_task/README.md
```diff
@@ -1,3 +1,42 @@
...
+{
+    "news_outlett_name": "news outlett name",
+    "news_title": "news title",
+    "pub_date": "publication date, Sun, 17 Nov 2019 10:44:20 -0500",
+    "news_link": "news link",
+    "news_description": "news description",
+    "img_alt": "image alternative if exist",
+    "img_src": "image link if exist",
+}
+```
```
HenadziStantchik: Please add information how to launch your app using setuptools

final_task/rss_reader/requirements.txt
```diff
@@ -1,14 +1,4 @@
-astroid==2.3.3
 beautifulsoup4==4.8.1
 bs4==0.0.1
```
HenadziStantchik: A question:

What is the difference between bs4 and beautifulsoup4? Do you really need both of them?

SergeyRogonov: My bad. It seems when you install like "pip install bs4" it installs beautifulsoup4 and this dummy package. I will remove it.

final_task/README.md
SergeyRogonov: I actually have a problem with it. To launch from source you can simply type "rss_reader.py [-h] [--version] [--json] [--verbose] [--limit LIMIT] source". But for installation via setuptools I can't name entry_point console_script from setup.py the same name as my entry point rss_reader.py. With other names it works fine. For now it works like "rss_reader [-h] [--version] [--json] [--verbose] [--limit LIMIT] source". I suppose it isn't OK so can you give me any tips, please. Or should I leave it be and just add the information about how to launch the app using setuptools?

HenadziStantchik: Try to bypass this problem using `sys.path.append` if you have troubles when import does not work either for setuptools or launching from source

final_task/rss_reader/news_articles.py
```diff
@@ -24,5 +24,16 @@ def __repr__(self):
...
+        return (self.news_outlett_name, self.news_title, self.pub_date,
+                self.news_link, self.news_description, self.img_alt, self.img_src)
+
+    def __hash__(self):
+        return hash(self.__key())
+
+    def __eq__(self, other):
+        if isinstance(other, NewsArticle):
+            return self.__key() == other.__key()
+        return NotImplemented
```
HenadziStantchik: A question: Why do you need to return this here?

final_task/README.md
```diff
@@ -17,14 +17,21 @@ optional arguments:
   --json         Print result as JSON in stdout
   --verbose      Outputs verbose status messages
   --limit LIMIT  Limit news topics if this parameter provided
+  --date         Read cached news for provided URL. If "ALL" provided - prints all cached news for this date
```
HenadziStantchik: Please add a bit more info about how do you store cached news. What do you use and etc.

HenadziStantchik:
I did not discover any issues with Iteration 3 work.

final_task/README.md
SergeyRogonov: Sure. Will fix it.

final_task/rss_reader/news_articles.py
SergeyRogonov: Initially I find it on stackoverflow as a good practice to implement __hash__(). In docs I found out that "If class does not define an __eq__() method it should not define a __hash__() operation either". So I decided to implement __hash__() with __eq__() together. And about "return NotImplemented" I found out that if the other object, we are comparing with, is not an instance of NewsArticle class, then this method goes back to built-in behavior.
Actually, for my project it is not necessary, so is it better I remove it? 
p.s. Do I get it all right?

final_task/rss_reader/requirements.txt
```diff
@@ -0,0 +1,21 @@
...
+nose2==0.9.1
+Pillow==6.2.1
+pycparser==2.19
+Pyphen==0.9.5
+python-dateutil==2.8.1
+six==1.13.0
+soupsieve==1.9.5
+tinycss2==1.0.2
+WeasyPrint==50
+webencodings==0.5.1
```
HenadziStantchik: There is too much redundant libs here. Usually only libs that are needed for your application _directly_ specified in this file.

final_task/rss_reader/cache.py
```diff
@@ -0,0 +1,105 @@
...
+                for article in cache[outlett]:
+                    if len(news_articles) == limit:
+                        break
+                    news_articles.append(article)
+        else:
+            # get news only for these url and date
+            if cache.get(url):
+                for article in cache[url]:
+                    if len(news_articles) == limit:
+                        break
```
HenadziStantchik: It is better to use slices (`[:]`) for this instead of comparing articles length to the limit on each iteration.

HenadziStantchik:
Iteration 1:
Does not work using `python rss_reader.py ....`
image

Iteration 2:
Installs, but does not work:
image

Iteration 3:
Could not be tested.

Iteration 4:
Could not be tested.

Tests are decently written, coverage is high.
The code is decently structured, readable, and absolutely correspond to the PEP8 guidelines.
Commit messages are informative and correspond to the recommended commit message style(good job)
Using templates for html is a plus.
Unfortunately application is not in working condition. It was tested both on a clean machine and inside isolated docker container

final_task/rss_reader/requirements.txt
SergeyRogonov: Oh, I see. I should not list all of them here even if most of them were installed by libs that I use directly?

HenadziStantchik: >Oh, I see. I should not list all of them here even if most of them were installed by libs that I use directly?

Yes. The same for `dependencies` in `setup.py`

SergeyRogonov: I get it. Thanks

final_task/rss_reader/cache.py
SergeyRogonov: Thanks. I fixed it.

PR #22, opened 2019-11-14
Oleg_Mikus_mikus.oleg@gmail.com

OlegMikus:


final_task/rss_reader/json_date_util.py
```diff
@@ -0,0 +1,26 @@
+import json
+
+
+def json_convert(dict_with_info: dict, limit: str):
```
HenadziStantchik: (Optional) If you are using typehints it is better to specify return type too.

final_task/rss_reader/json_date_util.py
```diff
@@ -0,0 +1,26 @@
...
+              'Entries': []
+              }
+    temp = {'Title': [],
+            'Date': [],
+            'Links': [],
+            'Info': []
+            }
+    if type(dict_with_info) == dict:
+        output['Feed'] = dict_with_info['Feed']
+        for index in range(int(len(limit))):
```
HenadziStantchik: During our sessions throughout the course we *emphasized multiple times* that you should __not__ use such constructions. Please fix it.

final_task/rss_reader/json_date_util.py
```diff
@@ -0,0 +1,26 @@
...
+
+    output = {'Feed': [],
+              'Entries': []
+              }
+    temp = {'Title': [],
+            'Date': [],
+            'Links': [],
+            'Info': []
+            }
+    if type(dict_with_info) == dict:
```
HenadziStantchik, line 16: It is better to use `isinstance` ~~of~~ `is` operator for type checks.

final_task/rss_reader/text_work_util.py
```diff
@@ -0,0 +1,92 @@
+import feedparser
+import html
+import logging
+
+
+def take_info(source: str, limit: str):
+    # take info from URL and create dict to storage news
```
HenadziStantchik: Here and below: Docstrings should be ike this: `"""Text"""`

final_task/rss_reader/text_work_util.py
```diff
@@ -0,0 +1,92 @@
...
+        for index in range(int(limit)):
+            print('\n' + 'Title: ' + dict_whit_info['Title'][index])
+            print('Date: ' + dict_whit_info['Date'][index])
+            print('Link: ' + dict_whit_info['Links'][index])
+            print('Info: ' + dict_whit_info['Info'][index])
+    else:
+        print(dict_whit_info)
+
+
+def make_text_better(text_of_news_list: list):
```
HenadziStantchik: Here, above and below: It is better to use more specific function names
Same for variable names

HenadziStantchik:
Here is some feedback on your app:

1. if `--limit` is not specified user should get all available feed.

2. It is recommended to not throw out image links but have them in the news feed, this will be helpful in the future iterations.

3. `--json` works strange

4. if `--limit` is larger than feed then you get an exception with traceback in stdout

OlegMikus:
> Here is some feedback on your app:
> 
> 1. if `--limit` is not specified user should get all available feed.
> 2. It is recommended to not throw out image links but have them in the news feed, this will be helpful in the future iterations.
> 3. `--json` works strange <- What does it mean?
> 4. if `--limit` is larger than feed then you get an exception with traceback in stdout

OlegMikus:
> 3. `--json` works strange

What does it mean?

HenadziStantchik:
> > 1. `--json` works strange
> 
> What does it mean?

If the limit is not specified, it will print 1 news entry. if the limit is ~1000 it will print 2 entries. if the --limit is ~124123127 it will print three entries. Please check its work

final_task/rss_reader/json_date_util.py
HenadziStantchik: or*

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,54 @@
...
+    parser.add_argument("--verbose", action="store_true", help="Outputs verbose status messages")
+    parser.add_argument("--limit", help="Limit news topics if this parameter provided",)
+
+    args = parser.parse_args()
+
+    return args
+
+
+def main_func():
+    limit = parsed().limit
```
HenadziStantchik, line 21: you don't really needd to call this function every time:
```
args = parsed()
limit  = args.limit
.....
```

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,54 @@
+import argparse
+import json_date_util as util
+import text_work_util as dt
+
+
+def parsed():
```
HenadziStantchik, line 6: It is better to use variable and function names that contain unambiguous answer to the question: `What is this?`

final_task/rss_reader/text_work_util.py
```diff
@@ -0,0 +1,147 @@
...
+    if limit:
+
+        try:
+            if int(limit) > len(all_info.entries) or int(limit) < 0:
+                limit = int(limit)
+            else:
+                limit = all_info.entries[0:int(limit)]
+        except ValueError:
+            print('\n Should try only integer numbers')
+            sys.exit()
```
HenadziStantchik, line 17: Here and below: It is better to have only one exit point from the application. For example you can raise your custom exception here or reraise this exception and handle it in the main function

final_task/rss_reader/text_work_util.py
```diff
@@ -0,0 +1,147 @@
...
+
+    else:
+        limit = all_info.entries
+
+    try:
+        for enter in limit:
+            dict_with_info['Title'] += [enter.title]
+            dict_with_info['Date'] += [enter.published]
+            dict_with_info['Links'] += [enter.link]
+            dict_with_info['Info'] += [enter.summary]
```
HenadziStantchik, line 27: A question:

Do you really need to put objects inside lists here? Why not just simply use these objects?

`[enter.title]` and etc. 

UPD: if you want to add one element to an existing list it is better to use `your_list.append(new_element)`

HenadziStantchik:
1. `--json` option still does not work:

mikus

2. If the limit is larger than feed size user should get all available news entries.

3. if option `--version` is specified by user, your app should print out version and then shutdown, without printing any news.

PR #28, opened 2019-11-20
Pavel_Los_lospawel@yandex.ru

paxalos:
5 iterations


HenadziStantchik:
Please make your PR name correspond to the one specified in our README.md

final_task/README.md
```diff
@@ -1,3 +1,56 @@
...
+    new_description:longtext
+    image_links:longtext
+When using the --date argument, news is searched by date in database
+
+How to install application
+To install application you should have setuptools. Open cmd and enter 'pip install -U setuptools'.
+Using 'pyhton setup.py install' in cmd install application.
+Install requirements 'pip install -r requirements.txt'
+You are now ready to run the application. Use 'rss-reader [arguments]' to run it.
+Warning: If path to rss-reader is not in Path variable, use full path to file at running.
```
HenadziStantchik, line 68: Please add documentation about ~~how do you store news in cache (Iteration 3)~~ some common info about Iteration 4 and 5

final_task/rss_reader/parse_rss_functions.py
```diff
@@ -0,0 +1,81 @@
...
+
+def ckeck_internet():
+    """
+    Checks Internet connetction
+    """
+    try:
+        logging.info("checking Internet connection")
+        socket.setdefaulttimeout(5)
+        host = socket.gethostbyname("www.google.com")
+        s = socket.create_connection((host, 80), 2)
```
HenadziStantchik: Still it is better to avoid one-letter variable names even in such simple cases

final_task/rss_reader/parse_rss_functions.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+    :param source - RSS URL:
+    :param limit - Limit of viewing news:
+    :return - RSS display list:
+    Function parsing the rss received from source
+    into a list of news which will then be used for printing or parsing into JSON
+    """
+    logging.info('Creating news list')
+    if not ckeck_internet():
+        raise NoInternet
```
HenadziStantchik: Here and below: It is better to raise exception insances, not objects, e.g. : `rainse NoInternetException("message")`

final_task/rss_reader/parse_rss_functions.py
```diff
@@ -0,0 +1,81 @@
...
+    logging.info('Getting and parsing RSS')
+    parsed_rss = feedparser.parse(source)
+    if parsed_rss['bozo'] == 1:
```
HenadziStantchik: `bozo` attribute is ether 1 or 0, so it is actually better to use `if parsed_rss['bozo']` here

final_task/rss_reader/personal_exceptions.py
```diff
@@ -0,0 +1,15 @@
+class IncorrectURL(Exception):
+    def __str__(self) -> str:
+        return 'The entered URL is incorrect'
```
HenadziStantchik: Here and below: You do not really need this logic if you will raise exceptions like this: `raise IncorrectUrl("message")`, `pass` will do

final_task/rss_reader/personal_exceptions.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class NoInternet(Exception):
+    def __str__(self) -> str:
+        return "Internet off, please check your connection"
+
+class IncorrectFilePath(Exception):
+    def __init__(self, message):
+        self.message=message
```
HenadziStantchik: A simple `pass` here will do.

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,112 @@
+import argparse
+import re
+from termcolor import colored
+import os
+import colorama
+from database_functions import *
```
HenadziStantchik: it is better to not use wildcard import at all

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,112 @@
...
+    argument_parser.add_argument('-v', '--verbose', action='store_true', help='increase output verbosity')
+    argument_parser.add_argument('--json', action='store_true', help='print result as JSON in stdout')
+    argument_parser.add_argument('--version', action='store_true', help='print version info')
+    argument_parser.add_argument('--limit', help='limit news topics if this parameter provided')
+    argument_parser.add_argument('--date', help='represent news from local storage by date')
+    argument_parser.add_argument('--to-html', help='save news in html format')
+    argument_parser.add_argument('--to-fb2', help='save news in fb2 format')
+    argument_parser.add_argument('--colorize', action='store_true', help='print news in colorized mode')
+    argument_parser.add_argument('source')
+    arguments = argument_parser.parse_args()
```
HenadziStantchik: It is better to move this to a separate function

HenadziStantchik:

1. User should be able to use `--version` without specifying RSS url

2. Your app does not work on Ubuntu OS:

los

paxalos:
Doesn't work, because you need to install mysql, then database final_task_database and table nes_cache like in readme

HenadziStantchik:
So the user should install MySQL client, create _specific database schema_ **and** _create specific table inside_ **by himself** just to be able to run your application?

How does this work with setuptools?

paxalos:
Yes

paxalos:
I used database, because it will be necessary in 6 iteration. But now I think that don't have time to do it, that is why I can change datebase storage  to store data in simple file. The database doesn't depend on setuptools

HenadziStantchik:
It is just not a good practice then. An average user (who does not know MySQL at all) will not be able to use your application.

It is up to you to decide what you are going to do, you still have 4 full days until the deadline.

If you really using database for Iteration 6 then it is ok, but make sure that steps of schema and table creation will be automated, so the user will just need to write 1-2 lines in the console to get things going on Iteration 6. You also can try some libs that use mysql, but actually do not require MySQL server installed and operational.

If you are not going to do the 6th iteration the it is of course preferable to simplify things for user: either try to automate it using `bash` scripts or just switching to a different storage method.

Or you can leave it as is. Of course it is not a good practice to make user manually setup the database, but we will surely **not** deny/fail your hard work just because of this issue.

P.S. I will test your app today a bit later

final_task/README.md
```diff
@@ -1,3 +1,56 @@
...
+Local storage
+All read news is saved in database by using Mysql. You should have database final_task_database with table news_cache
+new-cache structure:
+    feed:longtext
+    title:longtext
+    date:date
+    link:longtext
+    image_description:longtext
+    new_description:longtext
+    image_links:longtext
```
HenadziStantchik, line 48: Are you sure this is correct table structure?

I have set everything up, but I get warnings that date format is invalid and exception when trying to write news to the database:

los

HenadziStantchik:
Please check my comment on your README.md

final_task/README.md
paxalos: Sorry, my mistake. Format of date must be Datetime.

HenadziStantchik:
1. User should be able to user `--version` without specifying rss url

2. `--date` does not seem to work. I get "No news by this date` for any date I specify. ( while the table `news_cache` contains news data)

3. Currently you do not have any exception handling on actions with the databse. If anything goes wrong user gets hyge traceback to stdout, which shoiuld not happen

final_task/README.md
paxalos: Sorry again, date must have format date. I found mistake in write_news_to_database function, now I am fixing it

HenadziStantchik: Please fix it, and test it. I will look your pr once again  you will push changesa.

final_task/rss_reader/database_functions.py
```diff
@@ -1,7 +1,7 @@
 from contextlib import closing
 import logging
 import pymysql
-from dateutil import parser as date_parser
+from personal_exceptions import DatabaseConnectionError
```
HenadziStantchik: It is better to rename it to `custom_exceptions` of `errors`

final_task/rss_reader/database_functions.py
```diff
@@ -0,0 +1,73 @@
...
+                                     database='final_task_database')) as connection:
+            with closing(connection.cursor()) as cursor:
+                logging.info('Connected to database')
+                for new in news_list:
+                    # Try to find new in database by link, if exists
+                    cursor.execute(f'select * from news_cache where link = "{new["Link"]}"')
+                    if cursor.fetchall():
+                        continue
+                    insert_values = [value for value in new.values()]
+                    insert_values[6] = '|||'.join(insert_values[6])
```
HenadziStantchik, line 66: 6 is an oddly specific number, please add a comment here why 6 exactly

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,123 @@
...
+            if arguments.json:
+                print_news_JSON_colorize(news_list)
+            else:
+                print_news_colorize(news_list)
+        else:
+            if arguments.json:
+                print_news_JSON(news_list)
+            else:
+                print_news(news_list)
+    write_news_to_database(news_list)
```
HenadziStantchik: It is better to split this function a bit, for example you can move each argument check to a separate function: `check_html`, `check_fb2` and etc

HenadziStantchik:
1. I get constant warnings about incorrect date format. It is better either suppress them(not that good option) or fix your `datetime->date` conversion(preferable option)

2. `--date` does not work together with `--json`

3. HTML conversion does not work: if user specifies folder then `IsADirectoryError` is thrown with traceback in stdout. If user specifies file he gets `Incorrect html filepath` (Ubuntu OS)

Argument that user specifies for conversion to another format should be a directory, not a specific file. Moreover right now if you specify a directory instead of filename it will throw an exception with traceback in stdout

3. FB2 conversion does not work:

los

4. Keep in mind that conversion, `--colorize` and `--json` should work with `--date`

5. (optional) please rethink your color schema design for text background:
los

final_task/rss_reader/rss_reader.py
paxalos: Don't understand how it can be splitted

HenadziStantchik: for example 
```if arguments.to_html:
       check_html_arg(arguments.to_html)
   if arguments....
```

HenadziStantchik: for example:
```
if arguments.to_html:
    check_html_arg(arguments.html)
if arguments.....
```

HenadziStantchik:
There are some redundant symbols in fb2 and html converted files:
los

final_task/rss_reader/save_in_format_functions.py
```diff
@@ -47,7 +47,7 @@ def save_in_html(path, news_list):
     for new in news_list:
         html_content += get_new_content_html(new)
     html_content += "</body>\n</html>"
-    with open(path, 'w', encoding="utf-8") as html_file:
+    with open(os.path.join(path, "news.html"), 'w', encoding="utf-8") as html_file:
```
HenadziStantchik: It is better to have more informative name for news file. For example date-time-source and etc.
Same for fb2 format

paxalos: Can't find this reduntant symbols. Check your encoding, it must be utf8.

HenadziStantchik: Please try to parse the whole feed to the html and use Firefox browser to check it. (Linux OS)

paxalos: I got all news from yahoo.com to fb2, html and open them in notepad++, opera and firefox and don't met this symbols anywhere

HenadziStantchik:
I did not discover issues with Iteration 5 functionality. Still be sure to test it yourself, in case I missed something.

final_task/rss_reader/parse_rss_functions.py
```diff
@@ -0,0 +1,81 @@
...
+                          'Date':
+                              str(date_parser.parse(parsed_rss['entries'][index]['published'])).split(" ")[0],
+                          'Link':
+                              parsed_rss['entries'][index]['link'],
+                          'Image description':
+                              html.unescape(get_image_description(parsed_rss['entries'][index]['summary'])),
+                          'New description':
+                              html.unescape(get_new_description(parsed_rss['entries'][index]['summary'])),
+                          'Image links':
+                              [content['url'] for content in parsed_rss['entries'][index]['media_content']]})
```
HenadziStantchik, line 80: For multiline dictionary declarations it is better to use this codestyle:
```
news_dict = {
    "key`": value1,
    ....
}
```

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,101 @@
+import argparse
+import re
+from termcolor import colored
+import os
+import colorama
+import logging
+from database_functions import get_news_list_by_date, write_news_to_database
+from parse_rss_functions import get_news_list
+from custom_exceptions import NoInternet, IncorrectURL, IncorrectFilePath, DatabaseConnectionError
+from print_functions import print_news_colorize, print_news_JSON_colorize, print_news, print_news_JSON
```
HenadziStantchik, line 10: Here and everywhere else:
If you are using all functions inside imported module it is preferable to use just `import`

HenadziStantchik:
Here is some feedback on your application:
Iteration 1:
Everything works fine.
Iteration 2:
Everything installs and works, but only if you install libs specified in `requirements.txt` first, because not all of them are specified in the dependencies in `setup.py`
Iteration 3:
News caching does not work for news using Cyrillic symbols. (for example news.tut.by/rss)
Iteration 4:
Everything works fine.
Iteration 5:
Everything works fine.

Commit messages are informative enough and correspond to the recommended commit message style.
Tests are almost nonexistent and coverage percentage is not enough.
Code is mostly readable, but there are some large functions containing too much if/else branching, which lower readability. The code absolutely corresponds to the PEP8 guidelines.

final_task/README.md
```diff
@@ -1,3 +1,68 @@
...
+
+-h, --help            show this help message and exit
+--source SOURCE            RSS URL
+--version            Print version info
+--json            Print result as JSON in stdout
+--verbose            Outputs verbose status messages
+--limit LIMIT            Limit news topics if this parameter provided
+--date DATE            News from the specified day will be printed out. Format: YYYYMMDD
+It is mandatory to specify date or/and time.
+If both are specified, then news will be searched by date and by source.
```
HenadziStantchik, line 20: And what about Iteration 4 options?

paxalos: Forgot it

PR #30, opened 2019-11-23
Anastasiya_Merkushova_st.merkush@gmail.com

clonder:
Iteration 1-5 done


dzhigailo:
image
This commit messages are unacceptable.

Entire commit history doesn't fit standarts.
Solution is not covered with tests with is part one of the project's tasks.

PR #31, opened 2019-11-24
Anastasiya_Merkushova_st.merkush@gmail.com - PLS REVIEW

clonder:
1-5 done (without test), fix some problems with commits.


final_task/rss_reader/DatabaseEmulation.py
```diff
@@ -0,0 +1,93 @@
...
+    def __call__(cls, *args, **kwargs):
+        if cls not in cls._instances:
+            cls._instances[cls] = super(Singleton, cls).__call__(
+                *args,
+                **kwargs
+            )
+        return cls._instances[cls]
+
+
+class DatabaseEmulation(metaclass=Singleton):
```
dzhigailo: What is the reason to implement this class as Singleton? You create it only once anyway.

final_task/rss_reader/DatabaseEmulation.py
```diff
@@ -0,0 +1,93 @@
...
+            for line in data_file:
+                if not line:
+                    continue
+                date, row_number = line.strip().split()
+                self.dates[date] = int(row_number)
+        self.dates_file = open(path_to_dates, "a")
+        self.write_data_stream = open(self.path_to_data, "ab")
+        self.read_data_stream = open(self.path_to_data, "rb")
+
+    def __exit__(self, *args):
```
dzhigailo: This magic method doesn't make any sense without `__enter__(self)` magic method since both of those method are required for context manager protocol. Moreover, you never call it and opened files will not be closed.
Magic method __exit__ always have 3 positional arguments. There is no reason to use *args.

final_task/rss_reader/DatabaseEmulation.py
```diff
@@ -0,0 +1,93 @@
...
+    def write_items(self, rss_items: dict) -> None:
+        idx = len(self.dates)
+        date = now().strftime("%d/%m/%Y")
+        self.write_data_stream.seek(0, os.SEEK_END)
+        pickle.dump(rss_items, self.write_data_stream)
+        self.dates[date] = idx
+        self.dates_file.write("{} {}\n".format(date, idx))
+
+
+def get_txt_date(data) -> None:
```
dzhigailo: Function called `get_txt_date`, but it doesn't return anything.

final_task/rss_reader/DatabaseEmulation.py
```diff
@@ -0,0 +1,93 @@
+import pickle
```
dzhigailo: Module name doesn't not follow PEP standard. It should be in lower case.

final_task/rss_reader/DatabaseEmulation.py
```diff
@@ -0,0 +1,93 @@
...
+        date = now().strftime("%d/%m/%Y")
+        self.write_data_stream.seek(0, os.SEEK_END)
+        pickle.dump(rss_items, self.write_data_stream)
+        self.dates[date] = idx
+        self.dates_file.write("{} {}\n".format(date, idx))
+
+
+def get_txt_date(data) -> None:
+    date_id = {}
+    with open("dates.txt", "r", encoding="utf-8") as f:
```
dzhigailo: ` "r", encoding="utf-8"` these are default parameters for open() function.

dzhigailo: Same for code below.

final_task/rss_reader/converting.py
```diff
@@ -0,0 +1,96 @@
...
+        pdf.write(8, "Date: " + str(item["Date:"]))
+        pdf.ln(10)
+        img_url = str(item['Media content:\n'])
+        url_list = img_url.split("\n")
+        if check.internet_on and img_url != '':
+            for element in url_list:
+                if element.endswith(".png") or element.endswith(".jpg"):
+                    image = r.get(element)
+                    try:
+                        img_path = str(len(element)) + str(random.randint(1, 128)) + str(img_url[-4:])
```
dzhigailo: Here and below: you can't use random in path generation.

final_task/rss_reader/printers.py
```diff
@@ -0,0 +1,111 @@
...
+    news_link = item_xml['link']
+    description = split_string_by_lines(
+        filters.description(item_xml['description']),
+        10
+    )
+    media_content = split_string_by_lines(
+        ' '.join(item_xml['content']),
+        1
+    )
+    prepared_news = OrderedDict()
```
dzhigailo, line 42: In python version 3.7+ dictionary is ordered by default.

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,77 @@
...
+    log.info('try to read log file')
+    with open('logging.log', 'r', encoding="utf-8") as f:
+        print(f.read())
+
+
+if __name__ == '__main__':
+
+    log.info("Start script")
+
+    database = DatabaseEmulation.DatabaseEmulation(
```
dzhigailo: This class should support context manager protocol and use it.

dzhigailo:
First two commits should be squashed.
Tests are required by task definition.

final_task/rss_reader/DatabaseEmulation.py
clonder: It’s a good practice to make such objects as singleton. It doesn’t matter that in this case the object of this class is used once, it is much more important that if it is used a lot where, it was the same object.

final_task/rss_reader/printers.py
clonder: I know this, but if someone accidentally uses python3.6 for example, it will not be good if he has undefined behavior.

final_task/rss_reader/DatabaseEmulation.py
clonder: Fixed this, Thanks!

final_task/rss_reader/DatabaseEmulation.py
clonder: Sorry, missed this! Already fixed!

final_task/rss_reader/DatabaseEmulation.py
clonder: Fixed, thanks!

final_task/rss_reader/DatabaseEmulation.py
clonder: I didn’t know about this, thanks!
Fixed.

final_task/rss_reader/rss_reader.py
Clonder: Didn't catch this! Could you explain in more detail?

final_task/rss_reader/converting.py
clonder: Fixed!

final_task/rss_reader/requirements.txt
```diff
@@ -0,0 +1,28 @@
...
+numpy==1.17.4
+pycodestyle==2.5.0
+pycparser==2.19
+pyflakes==2.1.1
+pyparsing==2.4.5
+python-dateutil==2.8.1
+requests==2.22.0
+six==1.13.0
+soupsieve==1.9.5
+urllib3==1.25.7
```
HenadziStantchik, line 28: There are too many redundant libs here. Keep in mind that you should specify only libs that are _directly_ required by your application.

final_task/rss_reader/_database/dates.txt
```diff
@@ -0,0 +1,4 @@
+<<<<<<< HEAD
+=======
+
+>>>>>>> a9c6fd52b1a59d83145cc3e869cfe9f7a0a6c2fe
```
HenadziStantchik: You have unresolved conflicts here.

final_task/rss_reader/arguments.py
```diff
@@ -0,0 +1,55 @@
...
+
+version = '5.0'
+
+""" Add argument commands for script """
+
+
+def command_line():
+    parser = argparse.ArgumentParser(
+        description='Pure Python command-line RSS reader.',
+        prog='rss-reader')
```
HenadziStantchik, line 11: Here and below: it is better to make this into one line, it will be still pretty readable.
Also it is better to use `f""` instead of `.format()`.

final_task/rss_reader/converting.py
```diff
@@ -0,0 +1,146 @@
+import logging as log
+import requests as r
```
HenadziStantchik, line 2: It is better to avoid one-letter variable names.

final_task/rss_reader/converting.py
```diff
@@ -0,0 +1,146 @@
...
+            log.info("There is connection")
+            try:
+                for element in url_list:
+                    # Try to understand format of image file
+                    log.info("Try to understand format of image file")
+                    if element.endswith(".png") or element.endswith(".jpg"):
+                        image = r.get(element)
+                        try:
+                            img_path = str(len(element)) + \
+                                       str(img_url[-4:])
```
HenadziStantchik, line 90: It is better to use `os.path` for operations with filesystem paths because Windows OS and Linux OS have different separators

final_task/rss_reader/converting.py
```diff
@@ -0,0 +1,146 @@
...
+    result = create_html(items)
+    data = now().strftime("%d-%m-%Y")
+    time = now().strftime("%X")
+    name_of_html = str(data) + '_' + time
+    with open("{}".format(name_of_html), "w") as f:
+        f.write(str(result))
+        log.info("Successful converting into html")
+
+
+def create_pdf(items: list) -> None:
```
HenadziStantchik, line 45: This function is too large. It is better to split it.

final_task/rss_reader/databaseEmulation.py
```diff
@@ -0,0 +1,94 @@
...
+class Singleton(type):
+    _instances = {}
+
+    def __call__(cls, *args, **kwargs):
+        if cls not in cls._instances:
+            cls._instances[cls] = super(Singleton, cls).__call__(
+                *args,
+                **kwargs
+            )
+        return cls._instances[cls]
```
HenadziStantchik, line 21: It is better to not use such thing as singleton in this app. It is not needed at all.

final_task/rss_reader/databaseEmulation.py
```diff
@@ -0,0 +1,94 @@
...
+                date, row_number = line.strip().split()
+                self.dates[date] = int(row_number)
+        self.dates_file = open(path_to_dates, "a")
+        self.write_data_stream = open(self.path_to_data, "ab")
+        self.read_data_stream = open(self.path_to_data, "rb")
+
+    def __del__(self):
+        self.dates_file.close()
+        self.write_data_stream.close()
+        self.read_data_stream.close()
```
HenadziStantchik, line 42: It is better to use context manager's `__enter__` and `__exit__` for this purpose, not `__del__`

final_task/rss_reader/databaseEmulation.py
```diff
@@ -0,0 +1,94 @@
...
+
+    if data == date_id[key]:
+        date_id[key] = value
+        with open("{}.txt".format(value), "r") as f:
+            print(f.read())
+    else:
+        print("There isn't any news from this day")
+
+
+def cash_news() -> None:
```
HenadziStantchik, line 81: A typo here: `cache`

final_task/rss_reader/printers.py
```diff
@@ -0,0 +1,118 @@
...
+    prepared_news["Title:"] = title
+    prepared_news["Date:"] = date
+    prepared_news["Link: "] = news_link
+    prepared_news["Description: "] = description
+    prepared_news["Media content:\n"] = media_content
+    return prepared_news
+
+
+def filter_title(title: str) -> str:
+    new_title = title.replace("&#39;", '')
```
HenadziStantchik, line 53: It is better to have a bit more universal solution, as there might be other troublesome symbols.

final_task/rss_reader/printers.py
```diff
@@ -0,0 +1,118 @@
...
+
+def data_split(date: str) -> str:
+    log.info("Converting date")
+    month = {
+        'Dec': '12', 'Jan': '01', 'Feb': '02',
+        'Mar': '03', 'Apr': '04', 'May': '05',
+        'Jun': '06', 'Jul': '07', 'Aug': '08',
+        'Sep': '09', 'Oct': '10', 'Nov': '11'
+    }
+    new_date = date
```
HenadziStantchik, line 114: It is better to use `datetime` lib for this

HenadziStantchik:
It does not work right away
image


It does not work if installed using setuptools either (using `python setup.py install`):
image

Your app was tested both on clean machine and inside docker container.

The code is mostly readable, corresponds to the PEP8 guidelines.
Tests are almost nonexistent and coverage percentage is barely high enough.
Commits mostly are not informative enough and do not correspond to the recommended commit message style.

clonder:
Hi
1)I found a problem in the fact that I have a file dates in the folder _database in which for some reason there were strange lines with an error. (and it booted). If you delete them, my whole program works.
If there is such an opportunity - please look again. (I try it on two other machines)
2) for other screen I think you should try "python3 rss_reader.py -h" with _

PR #32, opened 2019-11-25
Denis_Marfonov_marfonovdenis@gmail.com

scarydzik:
Iteration 4 without unittests and docstrings

### DIFF
--- a/final_task/rss_reader/__main__.py
+++ b/final_task/rss_reader/__main__.py
@@ -1,0 +1,3 @@
+from rss_reader import main
+
+main()
--- a/final_task/rss_reader/__main__.py
+++ b/final_task/rss_reader/__main__.py
@@ -1,3 +1,3 @@
-from rss_reader import main
+from .rss_reader import main
 
 main()
--- a/final_task/rss_reader/book_gen.py
+++ b/final_task/rss_reader/book_gen.py
@@ -1,0 +1,110 @@
+import time
+import os.path
+import html
+
+from ebooklib import epub
+
+DATE_FORMAT = "%Y%m%d"
+
+
+def _render_document(title, items, date_str):
+    html = ["<html><head><title>"]
+    h_title = f"{date_str}"
+    html.append(h_title)
+    html.append("</title></head><body>")
+
+    html.append("<h1>")
+    html.append(title)
+    html.append("</h1>")
+
+    for i in items:
+        html.append(_render_item(i))
+
+    html = "".join(html)
+    return html
+
+
+def _render_item(item):
+    html = ["<hr>"]
+    title = item["title"] or "No Headline"
+    html.append(f"<h2>{title}</h2>")
+    date = time.strftime("%a, %d %b %Y %H:%M:%S", item["date"])
+    html.append(f"<p><i>{date}</i></p>")
+    if item["link"] is not None:
+        link = item["link"]
+        html.append(f"<p><a href='{link}'>{link}</a></p>")
+    else:
+        html.append(f"<p><a>no link</a></p>")
+    description = item["description"] or "No description"
+    html.append(f"<p>{description}</p>")
+    html = "".join(html)
+    return html
+
+
+def _gen_id(title: str, date:str) -> str:
+    """
+    generate string for book id and file name
+
+    :return: generated string
+    """
+    string = title + "_" + date
+    string = string.lower()
+    for c in r'\|/:*?"<>':
+        string.replace(c, '_')
+    if len(string) > 122:
+        string = string[:121] + "..."
+    return string
+
+
+def _create_html(book_id, bookpath, text):
+    # text = html.escape(text)
+    file = None
+    if os.path.isdir(bookpath):
+        file = open(os.path.join(bookpath, book_id + ".html"), "w", encoding="utf-8")
+    elif not os.path.exists(bookpath) or os.path.isfile(bookpath) or os.path.islink(bookpath):
+        file = open(bookpath, "w", encoding="utf-8")
+    try:
+        file.write(text)
+        file.close()
+    except (AttributeError, OSError):
+        pass
+
+
+def _create_epub(book_title, book_id, bookpath, text):
+
+    book = epub.EpubBook()
+    book.set_language('en')
+    book.set_identifier(book_id)
+    book.set_title(book_title)
+
+    content = epub.EpubHtml(title='News', file_name='content.xhtml')
+
+    content.set_content(text)
+    book.add_item(content)
+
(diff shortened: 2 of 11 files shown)


final_task/rss_reader/__main__.py
```diff
@@ -0,0 +1,3 @@
+from .rss_reader import main
```
HenadziStantchik, line 1: It is better to use simple `import` in this case

final_task/rss_reader/book_gen.py
```diff
@@ -0,0 +1,110 @@
...
+    html.append("</h1>")
+
+    for i in items:
+        html.append(_render_item(i))
+
+    html = "".join(html)
+    return html
+
+
+def _render_item(item):
```
HenadziStantchik, line 27: Here and below: Usually underscore should be used either inside classes for method names or for variables that override nuilt-in names. You do not actually need to use it every time.

final_task/rss_reader/book_gen.py
```diff
@@ -0,0 +1,110 @@
...
+    html = ["<html><head><title>"]
+    h_title = f"{date_str}"
+    html.append(h_title)
+    html.append("</title></head><body>")
+
+    html.append("<h1>")
+    html.append(title)
+    html.append("</h1>")
+
+    for i in items:
```
HenadziStantchik, line 20: Here and below, it is better to avoid one letter variable names. No matter how simple statement is.

final_task/rss_reader/book_gen.py
```diff
@@ -0,0 +1,110 @@
...
+    string = string.lower()
+    for c in r'\|/:*?"<>':
+        string.replace(c, '_')
+    if len(string) > 122:
+        string = string[:121] + "..."
+    return string
+
+
+def _create_html(book_id, bookpath, text):
+    # text = html.escape(text)
```
HenadziStantchik, line 60: It is better to avoid commented-out code in pushed commits.

final_task/rss_reader/book_gen.py
```diff
@@ -0,0 +1,110 @@
...
+    file = None
+    if os.path.isdir(bookpath):
+        file = open(os.path.join(bookpath, book_id + ".html"), "w", encoding="utf-8")
+    elif not os.path.exists(bookpath) or os.path.isfile(bookpath) or os.path.islink(bookpath):
+        file = open(bookpath, "w", encoding="utf-8")
+    try:
+        file.write(text)
+        file.close()
```
HenadziStantchik, line 68: It is better to use context manager for this instead of manually opening-closing file descriptor.

final_task/rss_reader/book_gen.py
```diff
@@ -0,0 +1,110 @@
...
+    except (AttributeError, OSError):
+        pass
```
HenadziStantchik, line 70: Usually if you leave `pass` in the `except` block it is better to leave a comment why.

final_task/rss_reader/feed.py
```diff
@@ -0,0 +1,207 @@
...
+class FeedNotFoundError(FeedError):
+    pass
+
+
+class IncorrectRSSError(FeedError):
+    pass
+
+
+class LocalCacheError(FeedError):
+    pass
```
HenadziStantchik, line 30: It is better to move your custom exceptions to the separate module

final_task/rss_reader/links_collection.py
```diff
@@ -0,0 +1,27 @@
...
+        return __x in self.items_dict
+
+    def __iter__(self):
+        return self.items_dict.__iter__()
+
+    def __init__(self, initial=None):
+        self.items_dict = dict()
+        self._next_index_ = 1
+
+        if initial is not None:
```
HenadziStantchik, line 19: It is better to be consistent with typehint usage: either typehint everywhere or nowhere.

final_task/rss_reader/rss_reader.py
```diff
@@ -0,0 +1,69 @@
+import argparse
+import logging
+import time
+import os.path
```
HenadziStantchik, line 4: Unused import here

final_task/rss_reader/book_gen.py
```diff
@@ -0,0 +1,110 @@
+import time
+import os.path
+import html
```
HenadziStantchik, line 3: Unused import here

HenadziStantchik:
Because you used `from .` in your imports, your app does not work if user will try to launch it using `python rss_reader.py <rss url> ...` (or something similar).
However, after package installation, your app still does not work (see Iteration 2 pic)
So in order to test your app I needed to change most of your imports.
Iteration 1:
After changing imports, does not work:
image
Iteration 2:
Installs nice, but does not work:
image

Iteration 3:
Unable to test.
Iteration 4:
Unable to test

Commits are informative and correspond to the recommended commit message style.
There are no tests.
Application is not in working condition, tested both on a clean machine and inside isolated docker container.
Overall code is readable, but has some issues specified in comments. It is however absolutely corresponds to the PEP8 guidelines.

scarydzik:
> Because you used from . in your imports, your app does not work if user will try to launch it using python rss_reader.py <rss url> ... (or something similar).

Program should be run with -m argument, as specified in the README:
```
$ python -m rss_reader ...
```
and from folder `final_tast/rss_reader` or from file `__main__.py`.
When I used absolute imports (`import foo` instead of `from . import foo`), console script from setup.py (rss-reader) didn't work.


As for the installation from PyPi, iteration 3 has version 0.3.post1 (because i accidently uploaded incompleted project as 0.3); iteration 4 was not added to TestPyPi, because I finished it just before deadline and forgot about it. Now I've added 4th iteration to TestPyPi.
Also it mandatory to install dependencies manualy. 4th iteration required `ebooklib`, which is not mentioned in README (cause I forgot), but is mentioned in requirements.txt and setup.py.

 If it's possible, can you please re-test the application?

HenadziStantchik:
Ok, my bad not reading your `README.md` attentively.

Your app still does not completely work though:

image

After a bit of investigating, I figured that the problem is that you are trying to connect/create to the DB file that is situated (in my case) in `final_task/rss_reader/data/` folder. The problem is that I do not have this folder at all, so it fails to open DB connection. However after creating `data` folder, everything works fine.

The same problem with your installed package

So here is feedback on your app:
Iteration 1:
`--json` contains html code, aside from that, everything is fine.
Iteration 2:
Installs fine, all iterations work the same as using `python -m ...`.
Iteration 3:
Everything works fine.
Iteration 4:
Html conversion works fine.
Epub conversion works fine, but contain no images, only placeholders for them:
image

